PR #20, opened 2019-11-18
Add Resource endpoint

jcoyne:
## Why was this change made?

So that the client can post the metadata somewhere.

## Was the API documentation (openapi.json) updated?
n/a


app/controllers/resources_controller.rb
```diff
@@ -0,0 +1,60 @@
...
+    Dor::Workflow::Client.new(url: Settings.workflow.url,
+                              logger: Rails.logger,
+                              timeout: 60)
+  end
+
+  # @return [Hash] the parameters used to register an object.
+  def register_params
+    reg_params = {
+      object_type: 'object',
+      admin_policy: 'TODO: what policy?'
```
justinlittman: ?

app/controllers/resources_controller.rb
```diff
@@ -0,0 +1,60 @@
...
+    }
+    reg_params[:label] = params[:collection_title].presence || ':auto'
```
justinlittman: What does `:auto` do / mean? A comment would be helpful.

app/controllers/resources_controller.rb
```diff
@@ -0,0 +1,60 @@
...
+      json: {
+        errors: [
+          {
+            "status": '504',
+            "title": msg,
+            "detail": err.message
+          }
+        ]
+      },
+      content_type: 'application/vnd.api+json',
```
justinlittman: Why not `application/json`?

app/controllers/resources_controller.rb
```diff
@@ -0,0 +1,60 @@
...
+      status: :gateway_timeout
```
justinlittman: Why is content_type and status going in the JSON?

app/controllers/resources_controller.rb
```diff
@@ -0,0 +1,60 @@
...
+    workflow_client.create_workflow_by_name(response[:pid], 'accessionWF')
+
+    render json: { druid: response[:pid] }, status: :created
+  rescue Dor::Services::Client::ConnectionFailed => e
+    render json: build_error('Unable to reach dor-services-app', e), status: :gateway_timeout
+  end
+
+  private
+
+  def workflow_client
```
justinlittman: Place in an initializer like dor services client?

spec/requests/create_resource_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+                "hasMember":["eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBOZz09IiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--89b7b484c80fe7f94d4aeff21c0c0e3e037d5c03"]
+              }
+            }
+          ]
+        }
+      } )
+    expect(JSON.parse(response.body)['druid']).to be_present
+    expect(response).to be_created
+  end
+end
```
justinlittman: Test error.

app/controllers/resources_controller.rb
jcoyne: I'm using this as a factory, not as a global initializer.   I think this is a better pattern.

app/controllers/resources_controller.rb
jcoyne: We don't have requirements for this yet, so placeholder.

app/controllers/resources_controller.rb
jcoyne: because this is  the registered mimetype for JSON api https://jsonapi.org/

app/controllers/resources_controller.rb
jcoyne: Yeah, that's not right, good catch.

app/controllers/resources_controller.rb
justinlittman: Cool. To help a dev not familiar with JSON-API, a link in the comment in L44 would be helpful to make it clear it is an external spec.

spec/requests/create_resource_spec.rb
jcoyne: Can you say more?

justinlittman: Test when a Dor::Services::Client::ConnectionFailed error occurs during `create`.

jcoyne: Thank you.

jcoyne: added.

PR #38, opened 2020-01-08
add rubocop and rake tasks 

### LINKED ISSUE #6: Add rubocop

ndushay:
## Why was this change made?

To add rubocop to this project

Fixes #6 

## Was the API documentation (openapi.json) updated?

n/a


Rakefile
```diff
@@ -6,3 +6,15 @@
 require_relative 'config/application'
 
 Rails.application.load_tasks
+
+begin
+  require 'rspec/core/rake_task'
+  require 'rubocop/rake_task'
+
+  desc 'Run linter and tests'
+  task ci: %i[spec rubocop]
```
mjgiarlo: Whaddya think about running `rubocop` before `spec` so we fail fast when there are style-related issues?

ndushay: sure, i can change that, np.

Issue #19, opened 2019-11-16
Add badges to readme


jcoyne:
- CI build
- coverage / maintainability

Issue #18, opened 2019-11-16
Add code coverage


jcoyne:
Preferring CodeClimate (over coveralls)

PR #44, opened 2020-01-08
add CI and coverage

### LINKED ISSUE #7: Add CI testing
there is a "ci" rake task, which is also the default.   But it does not currently spin up the dependencies needed for testing.

and we want to hook up with ci (circleci, prob.)

ndushay:
## Why was this change made?

Adds testing coverage and continuous integration.

Closes #7, #18 

## Was the API documentation (openapi.yml) updated?

n/a (badges will be added separately, see issue #19)

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -1,5 +1,3 @@
-# frozen_string_literal: true
-
 # This file is auto-generated from the current state of the database. Instead
 # of editing this file, please use the migrations feature of Active Record to
 # incrementally modify your database, and then regenerate this schema definition.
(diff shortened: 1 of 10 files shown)


db/schema.rb
```diff
@@ -12,38 +10,39 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20_191_115_013_418) do
+ActiveRecord::Schema.define(version: 2019_11_15_013418) do
+
   # These are extensions that must be enabled in order to support this database
-  enable_extension 'plpgsql'
+  enable_extension "plpgsql"
```
aaron-collier: Seems odd that we'd go from single to double quotes here (and throughout this file)

ndushay: db schema is auto-generated;  it should have been excluded from rubocop to begin with (my bad)

PR #50, opened 2020-01-09
Validate OpenAPI specification in CI

### LINKED ISSUE #41: validate openAPI as part of CI
~~blocked by #7~~

Loading the rails app (e.g. when running specs) automatically validates openAPI using  the `committee` gem, but the error messages are unhelpful in this context, so it will be better to have the CI build validate the openAPI.yml file separately, before running specs, in order to get more useful error messages.

Do this after rubocop, but before spec.  

This example for a json openapi spec, and validates the json itself after validating the openapi ... https://github.com/sul-dlss/dor-services-app/blob/master/.circleci/config.yml#L83-L101

We want to ensure it is valid yml BEFORE ensuring it is valid openapi.yml.

mjgiarlo:
Fixes #41

## Why was this change made?

When running CI, after linting the code, validate the API specification. Ensure it is valid per the OpenAPI Specification/OAS (semantic check).

Note that by validating OAS, we are also ensuring it is valid YAML (syntactic check). We do not need a separate tool or check for this.

This step is important since the API application uses the committee gem to parse the spec and enforce specified constraints, e.g., controller parameters, and the committee gem does not provide the most useful feedback when the spec is not valid.

Also, include the OpenAPI Spec validator badge in the README.

Yes.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2.1
 
 references:
   default_docker_ruby_executor: &default_docker_ruby_executor
-    image: circleci/ruby:2.6.4-stretch
+    image: circleci/ruby:2.6.4-stretch-node
     environment:
       BUNDLER_VERSION: 2.1.4
       BUNDLE_JOBS: 3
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -69,11 +69,30 @@ jobs:
...
-
       - run:
           name: Lint using rubocop
           command: bundle exec rubocop
-
+    - run:
+        name: Validate API specification
+        command: |
+          sudo npm install -g yaml-lint openapi-enforcer-cli
+          result=$(yamllint openapi.yml)
```
jcoyne: Do we need this part?  openapi-enforcer errors if it's not valid yaml.

mjgiarlo: No. Removed.

PR #61, opened 2020-01-09
Set the expectation that the file identifier is a string

### LINKED ISSUE #60: Bad request for /rails/active_storage/disk/{id}
The schema states this will be an integer, but it's a string. 
```
"{\"id\":\"bad_request\",\"message\":\"#/paths/~1rails~1active_storage~1disk~1{id}/put/parameters/0/schema expected integer, but received String: eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaDdDVG9JYTJWNVNTSWhhbkk1YlRNeFpuQnJkbmwxZG1zM01qYzFOM042ZURRMWRXYzBOQVk2QmtWVU9oRmpiMjUwWlc1MFgzUjVjR1ZKSWc1MFpYaDBMMmgwYld3R093WlVPaE5qYjI1MFpXNTBYMnhsYm1kMGFHa0Jrem9OWTJobFkydHpkVzFKSWgxemFURXZURE5xVEV0M05GRjVVRGw0UjNaMEsyRlJQVDBHT3daVSIsImV4cCI6IjIwMjAtMDEtMDlUMjE6MzI6NDguNDc1WiIsInB1ciI6ImJsb2JfdG9rZW4ifX0=--da785217f57ab55fe37e8de0a9c26b4d7612c187\"}"
```

jcoyne:
## Why was this change made?

Fixes #60

yes

### DIFF
--- a/openapi.yml
+++ b/openapi.yml
@@ -81,9 +81,9 @@ paths:
         - name: id
           in: path
           required: true
-          example: 123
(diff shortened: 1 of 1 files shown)


openapi.yml
```diff
@@ -83,7 +83,7 @@ paths:
           required: true
           example: 123
```
mjgiarlo: Looks like this may need to be made into a string.

jcoyne: Yeah, I guess so.

PR #66, opened 2020-01-10
Authorize create requests

jcoyne:
## Why was this change made?

Create requests should be made by a user.

n/a

### DIFF
--- a/spec/support/auth_helper.rb
+++ b/spec/support/auth_helper.rb
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+
+module AuthHelper
+  def jwt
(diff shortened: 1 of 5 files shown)


spec/support/auth_helper.rb
```diff
@@ -0,0 +1,15 @@
...
+
+module AuthHelper
+  def jwt
+    JsonWebToken.encode(payload)
+  end
+
+  private
+
+  def payload
+    user = User.create!(email: 'jcoyne85@stanford.edu', password: 'sekr3t!')
```
mjgiarlo: > pwned!!!!!!!!211

PR #67, opened 2020-01-10
Add an Ingest job

### LINKED ISSUE #17: Create ingest job
This should run in sidekiq. It will moves the files from the activestorage to the correct point in the `/dor/assembly` NFS mount, produces the correct XML metadata, and starts the assemblyWF.

Probably something like this: https://github.com/sul-dlss/pre-assembly/blob/master/app/lib/pre_assembly/digital_object.rb#L66-L70

jcoyne:
## Why was this change made?
Because the processing that needs to be done can take a long time.
Fixes #17

yes


app/jobs/ingest_job.rb
```diff
@@ -0,0 +1,23 @@
...
+    workflow_client.create_workflow_by_name(druid, 'accessionWF')
+    background_job_result.complete!
+  end
+
+  private
+
+  def workflow_client
+    Dor::Workflow::Client.new(url: Settings.workflow.url,
+                              logger: Rails.logger,
+                              timeout: 60)
```
mjgiarlo: Worth extracting to constant or configurable value, or meh?

app/jobs/ingest_job.rb
```diff
@@ -0,0 +1,23 @@
...
+class IngestJob < ApplicationJob
+  queue_as :default
+
+  def perform(druid:, background_job_result:)
+    background_job_result.processing!
+
+    # TODO: generate contentMetadata.xml
+    # TODO: ship files
+    workflow_client.create_workflow_by_name(druid, 'accessionWF')
+    background_job_result.complete!
```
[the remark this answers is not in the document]
jcoyne: Sure.

app/jobs/ingest_job.rb
jcoyne: very meh

mjgiarlo: I buy it.

app/jobs/ingest_job.rb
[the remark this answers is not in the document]
mjgiarlo: thx

PR #68, opened 2020-01-10
Install sidekiq for background job processing

jcoyne:
## Why was this change made?

So that we can use a queue for background job processing.

n/a

### DIFF
--- a/config/deploy.rb
+++ b/config/deploy.rb
@@ -25,6 +25,10 @@ append :linked_files, 'config/database.yml', 'config/honeybadger.yml', 'config/s
 # Default value for linked_dirs is []
 append :linked_dirs, 'log', 'config/settings', 'vendor/bundle'
 
+# Sidekiq configuration (run three processes)
+# see sidekiq.yml for concurrency and queue settings
+set :sidekiq_processes, 3
+
 # Default value for default_env is {}
 # set :default_env, { path: '/opt/ruby/bin:$PATH' }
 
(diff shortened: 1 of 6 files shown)


config/deploy.rb
```diff
@@ -25,6 +25,11 @@
 # Default value for linked_dirs is []
 append :linked_dirs, 'log', 'config/settings', 'vendor/bundle'
 
+# Sidekiq configuration (run three processes)
+# see sidekiq.yml for concurrency and queue settings
+set :sidekiq_processes, 3
+
+
```
mjgiarlo: Extra blank space. :keyboard:

jcoyne:
@mjgiarlo to what purpose?

mjgiarlo:
@jcoyne :speech_balloon: 
> @mjgiarlo to what purpose?

To spin up workers when running the app in containerland. I'm not sure it's needed, but we do seem to do this in some other projects, *e.g.*, https://github.com/sul-dlss/dor-services-app/blob/master/docker-compose.yml#L35-L52

jcoyne:
@mjgiarlo If it's in containerland, I'm assuming you are using development mode, so it'll be using async as the adapter, right?

mjgiarlo:
> @mjgiarlo If it's in containerland, I'm assuming you are using development mode, so it'll be using async as the adapter, right?

I see we're only using Sidekiq in the production env now, which is A-OK. No problem.

PR #69, opened 2020-01-10
Set logger to STDOUT on env variable

### LINKED ISSUE #22: Log to standard out
This complies with 12-factor methodology and opens the application to log aggregation. Add `lograge` which makes logs easier to parse.

aaron-collier:
## Why was this change made?

fixes #22

N/A - already set in docker-compose.yml

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -44,5 +44,7 @@ module SdrApi
 
     # This makes sure our Postgres enums function are persisted to the schema
     config.active_record.schema_format = :sql
+
+    config.logger = ActiveSupport::TaggedLogging.new(Logger.new(STDOUT)) if ENV['RAILS_LOG_TO_STDOUT'].present?
   end
 end
--- a/db/structure.sql
+++ b/db/structure.sql
@@ -22,6 +22,8 @@ CREATE TYPE public.background_job_result_status AS ENUM (
 
 SET default_tablespace = '';
 
+SET default_with_oids = false;
+
 --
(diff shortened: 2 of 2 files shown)


config/application.rb
```diff
@@ -44,5 +44,7 @@ class Application < Rails::Application
 
     # This makes sure our Postgres enums function are persisted to the schema
     config.active_record.schema_format = :sql
+
+    config.logger = ActiveSupport::TaggedLogging.new(Logger.new(STDOUT)) if ENV['RAILS_LOG_TO_STDOUT'].present?
```
mjgiarlo: If I'm reading this right, STDOUT will be used to log whether the `RAILS_LOG_TO_STDOUT` environment variable is set to true or false (because all keys and values in the `ENV` hash are string, AFAIK). I wonder if `ENV['RAILS_LOG_TO_STDOUT'].present?` should be something like `ENV['RAILS_LOG_TO_STDOUT'] == 'true'`?

db/structure.sql
```diff
@@ -22,6 +22,8 @@ CREATE TYPE public.background_job_result_status AS ENUM (
 
 SET default_tablespace = '';
 
+SET default_with_oids = false;
```
mjgiarlo: *was ist das?*

config/application.rb
jcoyne: This duplicates line 67 of config/environments/production.rb. I don't think we need it.  I think this ticket should just be a Puppet change.

aaron-collier: I was curious about that. I'll close this.

PR #73, opened 2020-01-10
Set a valid object_type for the registration request

### LINKED ISSUE #72: [sdr-api/stage] Dor::Services::Client::UnexpectedResponse: Bad Request: 400 (Unknown item type: 'object')
## Backtrace

    line 9 of [PROJECT_ROOT]/app/controllers/resources_controller.rb: create

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/67994/faults/59446521)

jcoyne:
## Why was this change made?
The previous value was not a valid option.
Fixes #72

n/a

### DIFF
--- a/app/controllers/resources_controller.rb
+++ b/app/controllers/resources_controller.rb
@@ -22,7 +22,7 @@ class ResourcesController < ApplicationController
   # @return [Hash] the parameters used to register an object.
   def register_params
     reg_params = {
-      object_type: 'object', # TODO: This can come from a lookup using params[:type]
+      object_type: 'item',
       admin_policy: params[:administrative][:hasAdminPolicy]
     }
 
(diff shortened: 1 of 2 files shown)


app/controllers/resources_controller.rb
```diff
@@ -33,6 +33,8 @@ def register_params
     reg_params[:metadata_source] = col_catkey ? 'label' : 'symphony'
     reg_params[:other_id] = "symphony:#{col_catkey}" if col_catkey
     reg_params[:collection] = params[:structural][:isMemberOf]
+    # TODO: content_type (tag Process:Content Type:Book (ltr) ) can come from a lookup using params[:type]
+    # reg_param[:tags] = ['Process:Content Type:Book (ltr)']
```
ndushay: is this already ticketed?

jcoyne: I'll do that now.

PR #80, opened 2020-01-13
Generate contentMetadata.xml for the uploaded files

### LINKED ISSUE #54: Create contentMetadata.xml
See https://github.com/sul-dlss-deprecated/dor-scripts/pull/31/files as an example.

jcoyne:
## Why was this change made?

So that the assembly WF has the required prerequisites.
Fixes #54

n/a

### DIFF
--- a/app/controllers/resources_controller.rb
+++ b/app/controllers/resources_controller.rb
@@ -15,7 +15,10 @@ class ResourcesController < ApplicationController
     end
 
     result = BackgroundJobResult.create
-    IngestJob.perform_later(druid: response[:pid], background_job_result: result)
+
+    IngestJob.perform_later(druid: response[:pid],
+                            filesets: params[:structural].to_unsafe_h.fetch(:hasMember),
+                            background_job_result: result)
 
     render json: { druid: response[:pid] },
            location: result,
--- a/app/jobs/ingest_job.rb
+++ b/app/jobs/ingest_job.rb
@@ -4,11 +4,20 @@
 class IngestJob < ApplicationJob
   queue_as :default
 
-  def perform(druid:, background_job_result:)
+  # @param [String] druid
+  # @param [Array<Hash>] filesets the data for creating the structure
+  # @param [BackgroundJobResult] background_job_result
+  def perform(druid:, filesets:, background_job_result:)
     background_job_result.processing!
 
-    # TODO: generate contentMetadata.xml
-    # TODO: ship files
+    dir = StagingDirectory.new(druid: druid, staging_location: Settings.staging_location)
+
+    file_names = copy_files_to_staging(dir, filesets)
+
+    # generate contentMetadata.xml
+    xml = ContentMetadataGenerator.generate(file_names: file_names, druid: druid)
+    dir.write_file('contentMetadata.xml', xml)
+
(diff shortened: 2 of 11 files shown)


app/jobs/ingest_job.rb
```diff
@@ -4,11 +4,20 @@
...
+  # @param [Array<Hash>] filesets the data for creating the structure
+  # @param [BackgroundJobResult] background_job_result
+  def perform(druid:, filesets:, background_job_result:)
     background_job_result.processing!
 
-    # TODO: generate contentMetadata.xml
-    # TODO: ship files
+    dir = StagingDirectory.new(druid: druid, staging_location: Settings.staging_location)
+
+    file_names = copy_files_to_staging(dir, filesets)
```
justinlittman: Should this copy be verified? (That is, verify checksums).

app/services/content_metadata_generator.rb
```diff
@@ -0,0 +1,71 @@
...
+
+  def cm_params
+    {
+      druid: druid,
+      objects: object_files,
+      file_attributes: file_attributes,
+      add_file_attributes: true,
+      add_exif: false,
+      bundle: :filename,
+      style: :simple_book
```
justinlittman: If this is a temporary hardcode, can you please call out with a comment?

app/controllers/resources_controller.rb
```diff
@@ -15,7 +15,10 @@ def create
     end
 
     result = BackgroundJobResult.create
-    IngestJob.perform_later(druid: response[:pid], background_job_result: result)
+
+    IngestJob.perform_later(druid: response[:pid],
```
justinlittman: I don't see a test for the IngestJob invocation.

app/jobs/ingest_job.rb
jcoyne: I don't think this is something we do when we do preassembly, so no.  If you feel strongly about it, open a ticket and we can discuss it with the PO.

app/controllers/resources_controller.rb
jcoyne: Okay, added one.

app/services/content_metadata_generator.rb
jcoyne: I don't know the answer to that without feedback from Andrew.

app/controllers/resources_controller.rb
mjgiarlo: I don't see one in the branch yet FWIW!

app/jobs/ingest_job.rb
justinlittman: https://github.com/sul-dlss/google-books/issues/110

jcoyne:
@mjgiarlo the code in StagingDirectory and ContentMetadataGenerator is well covered by the tests for IngestJob.  These classes support IngestJob in doing its work.  So the tests for IngestJob acts as an integration test for all of them.  I'm trying to follow the advice of Ken C. Dodds - "Write tests. Not too many. Mostly integration."

app/controllers/resources_controller.rb
jcoyne: oops, forgot to push.

PR #96, opened 2020-01-22
Add response for creating an object to openapi.yml.

### LINKED ISSUE #94: Document response structure for create a resource in openapi.yml
https://github.com/sul-dlss/sdr-api/blob/master/openapi.yml#L127-L129

justinlittman:
closes #94

## Why was this change made?
Document create object response.

Yes

### DIFF
--- a/openapi.yml
+++ b/openapi.yml
@@ -127,6 +127,16 @@ paths:
       responses:
         '200':
           description: OK
+          headers:
+            Location:
+              schema:
+                type: string
+                format: uri
+              description: URL to retrieve background job results.
+          content:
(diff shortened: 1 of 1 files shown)


openapi.yml
```diff
@@ -127,6 +127,10 @@ paths:
       responses:
         '200':
           description: OK
+          content:
```
jcoyne: Can you also add the Location header as well?

justinlittman: How does this look?

jcoyne: Love it.

PR #99, opened 2020-01-24
Allow passing md5 and sha1 for files and adds to content metadata.

justinlittman:
## Why was this change made?
To allow MD5 to be passed when depositing Google Books.

Yes, openapi.yml.

### DIFF
--- a/openapi.yml
+++ b/openapi.yml
@@ -302,6 +302,10 @@ components:
           type: string
         hasMimeType:
           type: string
+        hasMessageDigests:
+          type: array
+          items:
+            $ref: '#/components/schemas/MessageDigest'
         administrative:
           $ref: '#/components/schemas/FileAdministrative'
         access:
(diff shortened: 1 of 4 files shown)


openapi.yml
```diff
@@ -387,6 +391,22 @@ components:
           example: Searchworks
         release:
           type: boolean
+    MessageDigest:
+      description: The output of the message digest algorithm.
+      type: object
+      properties:
+        type:
```
jcoyne: Please add this (or ticket) to the cocina-models project.

openapi.yml
```diff
@@ -302,6 +302,10 @@ components:
           type: string
         hasMimeType:
           type: string
+        hasMessageDigests:
```
jcoyne: how about `hasFixity` (to follow premis) instead?

openapi.yml
```diff
@@ -387,6 +391,22 @@ components:
...
+      description: The output of the message digest algorithm.
+      type: object
+      properties:
+        type:
+          description: The algorithm that was used
+          type: string
+          enum:
+            - md5
+            - sha1
+        digest:
```
jcoyne: Can we call this `value` instead to conform to premis (circa-2017) which uses `rdf:value`?

justinlittman:
@jcoyne These were from cocina: https://github.com/sul-dlss/cocina-models/blob/master/docs/maps/File.json#L35-L52

PR #105, opened 2020-01-29
Sets use attribute on file if html file has hOCR element

### LINKED ISSUE #45: Add role=transcription for hOCR files
in the contentMetadata.

jermnelson:
## Why was this change made?
To set the use attribute in `contentMetadata.mxl` if a file is HTML and has an hOCR meta element.
Closes #45 

n/a

### DIFF
--- a/app/services/content_metadata_generator.rb
+++ b/app/services/content_metadata_generator.rb
@@ -59,6 +59,7 @@ class ContentMetadataGenerator
       file_node['publish'] = publish_attr(cocina_file)
       file_node['shelve'] = shelve_attr(cocina_file)
       file_node['preserve'] = preserve_attr(cocina_file)
+      ocr?(cocina_file) ? file_node['use'] = 'transcription' : ''
       cocina_file.fetch('hasMessageDigests', []).each do |message_digest|
         file_node.add_child(create_checksum_node(message_digest['type'], message_digest['digest']))
       end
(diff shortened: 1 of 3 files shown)


app/services/content_metadata_generator.rb
```diff
@@ -77,6 +78,17 @@ def preserve_attr(cocina_file)
     cocina_file.fetch('administrative').fetch('sdrPreserve') ? 'yes' : 'no'
   end
 
+  def ocr?(cocina_file)
+    if cocina_file.fetch('hasMimeType') == 'text/html'
+      # finds file in @file_names, parses HTML with Nokogiri for hOCR tags to return
+      # true if found
+      file_path = file_names.fetch(cocina_file.fetch('filename'))
+      doc = Nokogiri::HTML(File.open(file_path))
+      return true unless doc.css("meta[@name='ocr-capabilities']").empty?
```
jermnelson: Based on http://kba.cloud/hocr-spec/1.2/#metadata, the `ocr-capabilities` metadata element must be present for hOCR. We could also use `ocr-system` metadata element as well.

jermnelson:
> I think this might be better done in google-books. Then we can just send that use attribute without having to inspect the file contents.

Should issue #45 be transferred to Google books then? I can pick it up there.

jcoyne:
@jermnelson sure, go ahead.

jermnelson:
Issue moved to [sul-dlss/google-books #196](https://github.com/sul-dlss/google-books/issues/196) via [**ZenHub**](https://www.zenhub.com/)

jcoyne:
@jermnelson I still think we need a lot of this, we just should expect the source to be a cocina model attribute like 'preserve', 'shelve' and 'publish'.

PR #106, opened 2020-01-29
Remove outdated uploads on a schedule

### LINKED ISSUE #97: add a sweeper to remove uploads that are older than 7 days
Presumably these were deposits that failed after uploading files but before creating metadata.

mjgiarlo:
Fixes #97

## Why was this change made?

Includes:
* Add a new sweeper service that selects records to be swept based on an injectable strategy, defaulting to an empty set
* Add a null upload selection strategy and an outdated upload selection strategy
* Sweep uploads every day on a schedule
* Add configuration to allow setting a default "older than" date for determining which uploads to remove and a value to control ActiveRecord batch size for `AS::Blob`s
* Update cocina-models and dor-services-client (opportunistic upgrades)

No.

### DIFF
--- a/app/services/select_no_uploads_strategy.rb
+++ b/app/services/select_no_uploads_strategy.rb
@@ -1,0 +1,21 @@
+# frozen_string_literal: true
+
+# A strategy class that selects no uploads: a null strategy, if you will. Used
+# by the `DirectUploadsSweeper`
+class SelectNoUploadsStrategy
+  # @return [ActiveRecord::Relation]
+  def self.select
+    new.select
+  end
+
+  # @return [ActiveRecord::Relation]
+  def select
+    active_record_class.none
+  end
+
+  private
+
+  def active_record_class
+    ActiveStorage::Blob
+  end
+end
--- a/app/services/select_outdated_uploads_strategy.rb
+++ b/app/services/select_outdated_uploads_strategy.rb
@@ -1,0 +1,42 @@
+# frozen_string_literal: true
+
+# A strategy class that selects uploads older than a given or configured date.
+# Used by the `DirectUploadsSweeper`
+class SelectOutdatedUploadsStrategy
+  # @param [Integer] days_ago an integer representing the number of days before which uploads will be selected
+  # @return [ActiveRecord::Relation]
+  def self.select(days_ago: default_days_ago)
+    new(days_ago: days_ago).select
+  end
+
+  # @return [Integer]
+  def self.default_days_ago
+    Settings.sdr_api.days_after_which_to_remove_uploads
+  end
+
+  attr_reader :days_ago
+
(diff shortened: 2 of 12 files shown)


config/settings.yml
```diff
@@ -1,3 +1,7 @@
+sdr_api:
+  sweep_uploads_older_than: '7 days ago'
```
jcoyne: Can we make this an integer and get rid of the dependency on chronic?

app/services/select_no_uploads_strategy.rb
```diff
@@ -0,0 +1,21 @@
...
+
+  # @return [ActiveRecord::Relation]
+  def select
+    active_record_class.none
+  end
+
+  private
+
+  def active_record_class
+    @active_record_class ||= ActiveStorage::Blob
```
jcoyne: why an ivar here? There's no way to set this to anything else. Can this just return a const?

app/services/select_outdated_uploads_strategy.rb
```diff
@@ -0,0 +1,49 @@
...
+    # Parse to Time instance if not already passed a date/time-like object
+    Chronic.parse(date_time_representation)
+  end
+
+  def default_before_date
+    self.class.default_before_date
+  end
+
+  def active_record_class
+    @active_record_class ||= ActiveStorage::Blob
```
jcoyne: why ivar?

config/settings.yml
mjgiarlo: @jcoyne I pondered that, but... what would the integer mean? # of days? What if we want to switch to two weeks, for example? On one hand, I like being able to push changes like this by changing configuration alone, which wouldn't be possible if we switch this to an integer. On the other hand, changing days to weeks (or months or whatever) an easy code change.

app/services/select_no_uploads_strategy.rb
mjgiarlo: Not needed, will make the change you suggest.

app/services/select_outdated_uploads_strategy.rb
mjgiarlo: Not needed, will make the change you suggest.

app/services/select_no_uploads_strategy.rb
mjgiarlo: Done.

app/services/select_outdated_uploads_strategy.rb
mjgiarlo: Done.

config/settings.yml
mjgiarlo: @jcoyne FWIW, the `whenever` gem already adds `chronic` as a dependency. Still worth making this change?

mjgiarlo: done.

PR #111, opened 2020-01-31
Add a registered by tag (like Argo does)

jcoyne:
## Why was this change made?

So we can see who is using this service.

n/a

### DIFF
--- a/app/services/administrative_tags.rb
+++ b/app/services/administrative_tags.rb
@@ -2,13 +2,13 @@
 
 # Return a list of administrative tags for a given type URI
 class AdministrativeTags
-  def self.for(type:)
+  def self.for(type:, user:)
     case type
     when Cocina::Models::Vocab.book
       # NOTE: For now, we assume all books are LTR until we learn how to discern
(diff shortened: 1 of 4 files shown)


app/services/administrative_tags.rb
```diff
@@ -2,13 +2,13 @@
...
 class AdministrativeTags
-  def self.for(type:)
+  def self.for(type:, user:)
     case type
     when Cocina::Models::Vocab.book
       # NOTE: For now, we assume all books are LTR until we learn how to discern
       #       otherwise. See related issue:
       #       https://github.com/sul-dlss/google-books/issues/184
-      ['Process : Content Type : Book (ltr)'].freeze
+      ['Process : Content Type : Book (ltr)', "Registered By : #{user}"].freeze
```
mjgiarlo: I assume we'll want this tag for all work types, not just books, but I'm OK with this string living here for the time being.

jcoyne: yes, I figure we'll add more cases as the need arises.

PR #120, opened 2020-02-04
Adds use attribute for generating contentMetadata role

jermnelson:
## Why was this change made?
Adds the `use` attribute for generating the contentMetadata *role='transcription* for issue [#197](https://github.com/sul-dlss/google-books/issues/197) in Google Books. 

Yes


app/services/content_metadata_generator.rb
```diff
@@ -60,6 +60,7 @@ def create_file_node(id, cocina_file)
       file_node['publish'] = publish_attr(cocina_file)
       file_node['shelve'] = shelve_attr(cocina_file)
       file_node['preserve'] = preserve_attr(cocina_file)
+      cocina_file.key?('use') ? file_node['role'] = cocina_file.fetch('use') : nil
```
jcoyne:
```suggestion
      file_node['role'] = cocina_file.fetch('use') if cocina_file.key?('use')
```

jcoyne: Suggested this, because the `nil` value from the ternary wasn't used.

jermnelson: Done.

Issue #122, opened 2020-02-10
If embargoReleaseDate is set create an embargoMetadata datastream


jcoyne:
See https://github.com/sul-dlss/cocina-models/blob/master/lib/cocina/models/dro.rb#L35

Call this method: https://github.com/sul-dlss/dor-services/blob/master/lib/dor/models/concerns/embargoable.rb#L29-L31

justinlittman:
If embargoReleaseDate is set in what?

mjgiarlo:
@jcoyne Does this mean we need to add the `dor-services` gem to `sdr-api`, or that work should be done in `dor-services-app` (and `sdr-api` would integrate w/ it via a related change in `dor-services-client`? I would hope the latter.

jcoyne:
@mjgiarlo I don't know. I was just extracting this from sul-dlss/google-books#268

@justinlittman The deposit request in sdr-client should add this property here: https://github.com/sul-dlss/sdr-client/blob/master/lib/sdr_client/deposit/request.rb#L34 to follow the Cocina JSON schema: https://github.com/sul-dlss/cocina-models/blob/master/docs/maps/DRO.json#L89

justinlittman:
Embargo release date is already in the SDR API openapi.yml: https://github.com/sul-dlss/sdr-api/blob/master/openapi.yml#L196

PR #123, opened 2020-02-14
Switches to submitting cocina model when creating object.

justinlittman:
refs https://github.com/sul-dlss/sdr-api/issues/122

## Why was this change made?
So that SDR API creates an object using cocina model.

No.


mjgiarlo:
Can you take a look as well, @jcoyne ?

PR #133, opened 2020-02-20
Limited committee to API.

justinlittman:
## Why was this change made?
So that only the API is checked by committee.

Yes.

### DIFF
--- a/openapi.yml
+++ b/openapi.yml
@@ -24,40 +24,7 @@ tags:
     description: Digital Repository Objects
   - name: files
     description: upload binary files
-  - name: monitoring
-    description: Monitoring and health checks
 paths:
-  /status:
-    get:
-      tags:
-        - monitoring
-      summary: View default healthcheck information
-      responses:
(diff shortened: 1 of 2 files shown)


openapi.yml
```diff
@@ -24,40 +24,7 @@ tags:
     description: Digital Repository Objects
   - name: files
     description: upload binary files
-  - name: monitoring
-    description: Monitoring and health checks
 paths:
-  /status:
```
jcoyne: Isn't this an API though?

justinlittman: Sure, in some sense it is. But this is overhead that buys us nothing. Having to document OKComputer and Sidekiq and whatever else is a hassle and makes the openapi.yml even more verbose.

If there is a specific desire to expose status as part of the API, it should be given a `/v1/status` endpoint, as it is in dor-services-app.

jcoyne: Okay, that sounds reasonable.

PR #134, opened 2020-02-21
Adds creating registrationWF when registering.

justinlittman:
closes https://github.com/sul-dlss/google-books/issues/315

## Why was this change made?
registrationWF is run to create the registration milestone.

No.

### DIFF
--- a/app/controllers/resources_controller.rb
+++ b/app/controllers/resources_controller.rb
@@ -4,13 +4,24 @@ class ResourcesController < ApplicationController
   before_action :authorize_request
 
   # POST /objects
+  # rubocop:disable Metrics/AbcSize
+  # rubocop:disable Metrics/MethodLength
   def create
     begin
       response_cocina_obj = Dor::Services::Client.objects.register(params: cocina_model)
     rescue Dor::Services::Client::UnexpectedResponse => e
       return render build_error('Bad Request', e, '400', :bad_request)
     rescue Dor::Services::Client::ConnectionFailed => e
-      return render build_error('Unable to reach dor-services-app', e, '504', :gateway_timeout)
+      return render build_error('Unable to reach dor-services-app', e, '504',
+                                :gateway_timeout)
+    end
+
+    # Doing this here rather than IngestJob so that have accurate timestamp for milestone.
+    begin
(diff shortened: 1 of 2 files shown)


spec/requests/create_resource_spec.rb
```diff
@@ -132,20 +138,19 @@
           .to_return(status: 200, body: response_body, headers: {})
         # rubocop:enable Layout/LineLength
 
-        stub_request(:post, 'http://localhost:3001/objects/druid:abc123/workflows/accessionWF?lane-id=default')
```
justinlittman: This was cruft.

spec/requests/create_resource_spec.rb
```diff
@@ -168,28 +173,26 @@
...
 
         expect(response).to be_created
         expect(JSON.parse(response.body)['druid']).to be_present
         expect(IngestJob).to have_received(:perform_later)
+        expect(workflow_client).to have_received(:create_workflow_by_name).with('druid:abc123',
+                                                                                'registrationWF', version: 1)
       end
     end
 
-    context 'when the registration request is unsuccessful' do
```
justinlittman: Removed this test since it is a duplicate.

app/controllers/resources_controller.rb
```diff
@@ -4,13 +4,24 @@ class ResourcesController < ApplicationController
...
       return render build_error('Bad Request', e, '400', :bad_request)
     rescue Dor::Services::Client::ConnectionFailed => e
-      return render build_error('Unable to reach dor-services-app', e, '504', :gateway_timeout)
+      return render build_error('Unable to reach dor-services-app', e, '504',
+                                :gateway_timeout)
+    end
+
+    # Doing this here rather than IngestJob so that have accurate timestamp for milestone.
+    begin
+      workflow_client.create_workflow_by_name(response_cocina_obj.externalIdentifier, 'registrationWF', version: 1)
```
jcoyne: Should this be in dor-services-app in the create method?  Then this service won't need a dependency on the workflow service.

jcoyne: It turns out that argo doesn't apply registrationWF to all objects.  We can remove this line once https://github.com/sul-dlss/dor-services-app/issues/681 is complete.

PR #136, opened 2020-02-21
Use correct HTTP error code in JSON Error

### LINKED ISSUE #132: Error message inconsistency.
```json
{"errors":[{"status":"400","title":"Bad Request","detail":"Conflict: 409 (An object with the source ID 'googlebooks:stanford_20500274329' has already been registered.)"}]}
```

It's returning 400, but the detail says 409.  We should return a 409 in this case.

jermnelson:
## Why was this change made?
Fixes #132. 

n/a

### DIFF
--- a/app/controllers/resources_controller.rb
+++ b/app/controllers/resources_controller.rb
@@ -49,13 +49,15 @@ class ResourcesController < ApplicationController
     Cocina::Models::RequestDRO.new(model_params)
   end
 
-  # JSON-API error response. See https://jsonapi.org/
+  # JSON-API error response. See https://jsonapi.org/.
   def build_error(msg, err, code, status)
+    m = err.message.match(/:\s(\d{3})/)
+    !m.nil? && m[1] != code ? code = m[1] : ''
     {
       json: {
         errors: [
           {
-            "status": code !== err.status ? err.status : code,
+            "status": code,
             "title": msg,
             "detail": err.message
           }
--- a/app/controllers/resources_controller.rb
+++ b/app/controllers/resources_controller.rb
@@ -50,7 +50,7 @@ class ResourcesController < ApplicationController
   end
 
   # JSON-API error response. See https://jsonapi.org/.
-  def build_error(msg, err, code, status)
+  def build_error(msg, err, code, _status)
     m = err.message.match(/:\s(\d{3})/)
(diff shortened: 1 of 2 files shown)


app/controllers/resources_controller.rb
```diff
@@ -49,8 +49,10 @@ def cocina_model
     Cocina::Models::RequestDRO.new(model_params)
   end
 
-  # JSON-API error response. See https://jsonapi.org/
+  # JSON-API error response. See https://jsonapi.org/.
   def build_error(msg, err, code, status)
+    m = err.message.match(/:\s(\d{3})/)
```
jermnelson: Using a regex based on the return value in `dor-services-client` ResponseErrorFormatter. More involved solution would be to add the error code as a variable to `Dor::Services::Client::UnexpectedResponse`, `Dor::Services::Client::ConnectionFailed`, and maybe `Dor::WorkflowException`

spec/requests/create_resource_spec.rb
```diff
@@ -207,6 +207,24 @@
...
+                     "Conflict: 409 (An object with the source ID 'abcd:1234' already registered)")
+      end
+
+      let(:error) { JSON.parse(response.body)['errors'][0] }
+
+      it 'returns an error response' do
+        post '/v1/resources',
+             params: request,
+             headers: { 'Content-Type' => 'application/json', 'Authorization' => "Bearer #{jwt}" }
+        expect(response).to have_http_status(:bad_request)
```
jcoyne: This status should be conflict as well.

jermnelson: Do you mean the `:bad_request`? The intention (I think)  in the `build_error` method is to follow the https://jsonapi.org/format/#errors suggestions, the over-all response has a more general error of `400` because there could be multiple errors with different error codes in the JSON `error` array.

jcoyne: While there could be more than one code in the array according to JSON-API, this is never true in our case.  We just want this to act as a simple proxy to dor-service-app and return the same error code that sdr-api received.

jermnelson: 👍 Just wanted to confirm that assumption. Pushed commit that sets status to error code if it exists.

PR #138, opened 2020-02-25
Return errors from dor-services-app as 502 not 400

justinlittman:
## Why was this change made?
To return a more accurate error code when call to dor-services-app returns an error.

Yes.

### DIFF
--- a/app/controllers/resources_controller.rb
+++ b/app/controllers/resources_controller.rb
@@ -10,7 +10,7 @@ class ResourcesController < ApplicationController
     begin
       response_cocina_obj = Dor::Services::Client.objects.register(params: cocina_model)
     rescue Dor::Services::Client::UnexpectedResponse => e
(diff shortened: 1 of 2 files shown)


app/controllers/resources_controller.rb
```diff
@@ -10,7 +10,7 @@ def create
     begin
       response_cocina_obj = Dor::Services::Client.objects.register(params: cocina_model)
     rescue Dor::Services::Client::UnexpectedResponse => e
-      return render build_error('Bad Request', e, '400', :bad_request)
+      return render build_error('Error registering object with dor-services-app', e, '502', :bad_gateway)
```
ndushay: is this only called for registering objects?  (is the language correct)

justinlittman: Yes, given where the rescue is placed this is only called for registering objects.

PR #144, opened 2020-02-27
Don't build contentMetadata, the cocina-datamodel has it

jcoyne:
## Why was this change made?

dor-services-app is now responsible for building contentMetadata from the cocina data model.

n/a

If so, please confirm:
- [ ] change was tested on stage    and/or
- [ ] test added to sul-dlss/infrastructure-integration-test


jcoyne:
I'm not sure this makes sense to proceed with at this point because, if we skip assemblyWF, then the generated JP2s (for shelving) will not be included in the contentMetadata.

Should we should consider adding jp2 generation here?

jcoyne:
@ndushay I think the code coverage drop has more to do with removing covered code than from adding anything that isn't covered.

PR #158, opened 2020-03-04
Add a proxy request

jcoyne:
## Why was this change made?
So that argo can get the credentials to give back to the user.  This will allow them to make a sdr-api deposit.
This change will support integration tests of sdr-api

yes

If so, please confirm change was tested on stage and/or test added to sul-dlss/infrastructure-integration-test.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -126,9 +126,9 @@ GEM
       capistrano-one_time_key
       capistrano-shared_configs
     docile (1.3.2)
-    dor-services-client (4.18.0)
+    dor-services-client (4.19.0)
       activesupport (>= 4.2, < 7)
-      cocina-models (~> 0.28)
+      cocina-models (~> 0.29)
       deprecation
       faraday (>= 0.15, < 2)
       moab-versioning (~> 4.0)
@@ -201,7 +201,7 @@ GEM
     globalid (0.4.2)
       activesupport (>= 4.2.0)
     hashdiff (1.0.1)
-    honeybadger (4.5.6)
+    honeybadger (4.6.0)
     i18n (1.8.2)
       concurrent-ruby (~> 1.0)
     ice_nine (0.11.2)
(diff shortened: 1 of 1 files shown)


config/settings.yml
```diff
@@ -11,3 +11,8 @@ workflow:
   url: 'http://localhost:3001'
 
 staging_location: '/dor/workspace'
+
+
+# This user is allowed to proxy other users, we expect that it'll be argo and it
+# will be responible for only proxying users which it has verififed with Shibboleth.
+argo_user: argo@dlss.sul.stanford.edu
```
ndushay: this email should be clearly a dummy email (e.g. `overwritten@example.org`) or it should be a real one.  I would lean towards a real one, maybe sent to team list so if it gets used we become aware things aren't working right.

openapi.yml
```diff
@@ -155,6 +155,35 @@ paths:
...
+                  token:
+                    type: string
+                    description: The access token
+                    example: 'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiaWF0IjoxNTE2MjM5MDIyfQ.SflKxwRJSMeKKF2QT4fwpMeJf36POk6yJV_adQssw5c'
+                  exp:
+                    type: string
+                    description: When the token will expire
+                    format: date-time
+      parameters:
+        - in: query
```
mjgiarlo: Is it a query arg or is it sent in the response body?

jcoyne: It's a query arg.

config/settings.yml
jcoyne: @ndushay I need it to uniquely identify argo (and not other systems).  Which one would you suggest?

ndushay: @andrewjbtw is there an existing argo email that makes sense?  should we have one for stage also?   Or just use dummy email addresses?

ndushay: - will the email above EVER get any email?  if no, could be something like `proxying-user@argo-prod.stanford.edu`. If yes ... who should get the email?  should it be list you set up with mailman that's for devs + Andrew?

- do we need it to be different for prod/stage/local development?  If so, the comment should indicate and/or it should be in shared_configs or something.  If no, then `proxying-user-4-argo@dlss.sul.stanford.edu` or some such would be good.

jcoyne: no email is being sent here, it's just used to identify a service or a user.  It doesn't need to change in different environments.

ndushay: then cool.  do as you wish.

PR #162, opened 2020-03-11
Remove external identifier as file attribute in openapi.

justinlittman:
## Why was this change made?
External identifier is an attribute of a File not a Request File.

See https://github.com/sul-dlss/cocina-models/blob/master/lib/cocina/models/file.rb#L45

Yes.

If so, please confirm change was tested on stage and/or test added to sul-dlss/infrastructure-integration-test.
No.

### DIFF
--- a/openapi.yml
+++ b/openapi.yml
@@ -298,8 +298,6 @@ components:
           type: string
         filename:
           type: string
-        externalIdentifier:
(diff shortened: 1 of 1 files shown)


justinlittman:
Turns out that SDR API uses externalIdentifier.

PR #194, opened 2020-05-15
Moves all ingest steps to IngestJob.

justinlittman:
## Why was this change made?
So that google books deposits do not fail in bad states and make the service more resilient.

Previously, register and starting the registrationWF were performed synchronously in the resources controller. However, on a regular basis this would timeout, leaving the item as registered but the rest of the deposit steps incomplete.

This moves the entire deposit operation to a job that can be retried. Retries can handle if the deposit was only partially completed.

The client can receive updates on the deposit by checking the background job result endpoint.

It has been tested in stage.

None.


app/controllers/resources_controller.rb
```diff
@@ -6,78 +6,20 @@ class ResourcesController < ApplicationController
...
-      workflow_client.create_workflow_by_name(response_cocina_obj.externalIdentifier, 'registrationWF', version: 1)
-    rescue Dor::WorkflowException => e
-      return render build_error('Error creating registrationWF with workflow-service', e, '502')
-    end
-
-    result = BackgroundJobResult.create
-    IngestJob.perform_later(druid: response_cocina_obj.externalIdentifier,
-                            filesets: params[:structural].to_unsafe_h.fetch(:contains, []),
+    result = BackgroundJobResult.create(output: {})
+    IngestJob.perform_later(model_params: params.except(:action, :controller, :resource, :accession).to_unsafe_h,
```
jcoyne, line 12: How about we pass the cocina model, that way the consumer at least gets fast validation of the structure of what they sent and that all the files they reference are something sdr-api is aware of.

justinlittman: Won't this validation will already be provided by openapi / committee?

app/controllers/resources_controller.rb
```diff
@@ -6,78 +6,20 @@ class ResourcesController < ApplicationController
...
-    begin
-      workflow_client.create_workflow_by_name(response_cocina_obj.externalIdentifier, 'registrationWF', version: 1)
-    rescue Dor::WorkflowException => e
-      return render build_error('Error creating registrationWF with workflow-service', e, '502')
-    end
-
-    result = BackgroundJobResult.create
-    IngestJob.perform_later(druid: response_cocina_obj.externalIdentifier,
-                            filesets: params[:structural].to_unsafe_h.fetch(:contains, []),
+    result = BackgroundJobResult.create(output: {})
```
mjgiarlo, line 11: I wonder if it's worth tossing a simple test of this new interaction into the deleted request spec below.

app/controllers/resources_controller.rb
jcoyne: I guess the structural validation would, but not necessarily "we know about these" files validation.

app/jobs/ingest_job.rb
```diff
@@ -1,26 +1,60 @@
...
-    dir = StagingDirectory.new(druid: druid, staging_location: Settings.staging_location)
-    file_nodes = filesets.flat_map { |fs| fs.fetch('structural').fetch('contains') }
-    copy_files_to_staging(dir, file_nodes)
+    begin
+      response_cocina_obj = Dor::Services::Client.objects.register(params: cocina_model(model_params))
+      druid = response_cocina_obj.externalIdentifier
+    rescue Dor::Services::Client::ConflictResponse => e
+      # Should not expect this on first try so return as error
+      if background_job_result.try_count == 1
+        background_job_result.output = { errors: [title: 'Object with source_id already exists.', message: e.message] }
```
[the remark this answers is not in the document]
justinlittman: Yes, the error message containst the source_id and druid.

I'd suggest that it would be the client's responsibility to decide whether to HB.

app/jobs/ingest_job.rb
```diff
@@ -1,26 +1,60 @@
...
+    background_job_result.complete!
+  rescue StandardError => e
+    # This causes Sidekiq to retry.
+    if background_job_result.try_count < Settings.sdr_api.ingest_retries
+      background_job_result.pending!
+      raise
+    end
+
+    # Otherwise return an error on background_job_result but exit cleanly.
+    background_job_result.output.merge!({ errors: [title: 'All retries failed', message: e.message] })
```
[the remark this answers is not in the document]
justinlittman: If a druid was created, it would be included in the response (hence the `merge!`).

app/jobs/ingest_job.rb
```diff
@@ -42,7 +75,7 @@ def copy_files_to_staging(dir, file_nodes)
   # @param [Array<Hash>] a list of hashes representing Cocina File objects
   # @return [Array<ActiveStorage::Blob>] Given a list of file_json, return the corresponding blob objects
   def files(file_nodes)
-    signed_ids = file_nodes.map { |fs| fs.fetch('externalIdentifier') }
+    signed_ids = file_nodes.map { |fs| fs.fetch(:externalIdentifier) }
```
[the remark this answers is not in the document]
justinlittman: The keys are symbolized; I don't think it has indifferent access.

app/jobs/ingest_job.rb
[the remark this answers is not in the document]
mjgiarlo: @justinlittman :+1:

app/controllers/resources_controller.rb
justinlittman: Hmmm ... the request spec wasn't deleted. The diff was just to large to display. Is the spec OK or are you suggesting testing something additional?

mjgiarlo: @justinlittman mea culpa. Yep, that's totally sufficient.

app/controllers/resources_controller.rb
justinlittman: What do you mean about the "we know these" files validation? That we can find an ActiveStorage blob for each Cocina file? If so, we could, but doesn't seem necessary to do early validation of a problem that we've never had and will catch anyway.

app/controllers/resources_controller.rb
```diff
@@ -3,84 +3,85 @@
...
+    rescue BlobError => e
+      # Returning 500 because not clear whose fault it is.
+      return render build_error('500', e, 'Error matching uploading files to file parameters.')
     end
-
-    result = BackgroundJobResult.create
-    IngestJob.perform_later(druid: response_cocina_obj.externalIdentifier,
-                            filesets: params[:structural].to_unsafe_h.fetch(:contains, []),
+    result = BackgroundJobResult.create(output: {})
+    IngestJob.perform_later(model_params: JSON.parse(request_dro.to_json), # Needs to be sidekiq friendly serialization
```
jcoyne, line 21: I think this is fine, but you could have also done `model_params: request_dro.to_json`, right?

justinlittman: Yes, which would have passed a string instead of a hash.

PR #201, opened 2020-05-27
Ignore symphony

justinlittman:
## Why was this change made?
To account for nightly Symphony outages by:
* Not HB alerting.
* Retry enough to get pass outages.

To be tested on stage.

Yes.


mjgiarlo:
:clap:

PR #237, opened 2020-10-21
[HOLD] Updates cocina models and dor-services-client.

justinlittman:
## Why was this change made?
Update cocina models and dor-services-client.

NA

openapi

### DIFF
--- a/openapi.yml
+++ b/openapi.yml
@@ -203,15 +203,6 @@ components:
             - 'citation-only'
             - 'dark'
           default: 'dark'
-        download:
-          description: Download access level for a file
-          type: string
-          enum:
-            - 'world'
-            - 'stanford'
-            - 'location-based'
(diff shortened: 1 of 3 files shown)


openapi.yml
```diff
@@ -241,7 +232,7 @@ components:
         - hasAdminPolicy
     AdminPolicy:
       type: object
-      additionalProperties: false
+      additionalProperties: true
```
jcoyne, line 235: Can we get away with not accepting extra properties? I think this is a valuable safeguard against submitting incorrect data.

mjgiarlo (approved):

openapi.yml
justinlittman: I agree. However, in some cases it causes committee problems. If I recall correctly, it is in endpoints that have parameters.

PR #238, opened 2020-10-22
Set RequestDRO additionalProperties to true.

justinlittman:
## Why was this change made?
In order for committee to allow the `accession` query param, `additionalProperties` has to be set to true.

Integration

openapi

### DIFF
--- a/openapi.yml
+++ b/openapi.yml
@@ -1216,7 +1216,8 @@ components:
     RequestDRO:
       description: A request to create a DRO.  This has the same general structure as a DRO but doesn't have externalIdentifier and doesn't require the access subschema. If no access subschema is provided, these values will be inherited from the AdminPolicy.
       type: object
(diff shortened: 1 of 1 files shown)


openapi.yml
```diff
@@ -1216,7 +1216,7 @@ components:
     RequestDRO:
       description: A request to create a DRO.  This has the same general structure as a DRO but doesn't have externalIdentifier and doesn't require the access subschema. If no access subschema is provided, these values will be inherited from the AdminPolicy.
       type: object
-      additionalProperties: false
+      additionalProperties: true
```
jcoyne, line 1219: Would you mind adding a comment about that here?

jcoyne (approved):

openapi.yml
justinlittman: Yup. Done.

PR #245, opened 2020-11-17
Add an update API

jcoyne:
## Why was this change made?

This enables h2 to update the metadata.

### DIFF
--- a/app/controllers/resources_controller.rb
+++ b/app/controllers/resources_controller.rb
@@ -2,17 +2,18 @@
 
 require 'base64'
 
+# rubocop:disable Metrics/ClassLength
 class ResourcesController < ApplicationController
   class BlobError < StandardError; end
 
   before_action :authorize_request
 
-  # POST /objects
+  # POST /resource
   def create
     authorize! :resource
 
     begin
-      request_dro = cocina_model(params.except(:action, :controller, :resource, :accession).to_unsafe_h)
+      request_dro = cocina_request_model(params.except(:action, :controller, :resource, :accession).to_unsafe_h)
     rescue BlobError => e
       # Returning 500 because not clear whose fault it is.
       return render build_error('500', e, 'Error matching uploading files to file parameters.')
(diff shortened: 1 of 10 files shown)


app/controllers/resources_controller.rb
```diff
@@ -28,35 +29,80 @@ def create
...
     Cocina::Models::RequestDRO.new(new_model_params)
   end
 
   # Decorates the provided FileSets with the information we have in the ActiveStorage table.
   # externalIdentifier is also removed from the request.
-  # rubocop:disable Metrics/AbcSize
   def decorate_file_sets(model_params)
     file_sets(model_params).each do |fileset|
-      fileset[:version] = 1
+      fileset[:version] = model_params[:version]
```
justinlittman, line 73: Fileset version changes even if files haven't?

app/controllers/resources_controller.rb
```diff
@@ -28,35 +29,80 @@ def create
...
+                      version: model_params[:version])
       end
     end
   end
+
+  # rubocop:disable Metrics/AbcSize
+  def decorate_file(file:, version:, external_id: nil)
+    blob = blob_for_signed_id(file.delete(:externalIdentifier), file[:filename])
+    file[:externalIdentifier] = external_id if external_id
+    file[:version] = version
```
justinlittman, line 86: File version changes even if file hasn't?

app/controllers/resources_controller.rb
jcoyne: Yes, this is synched to the parent, since we don't track file versions separately in Fedora. I think this could change when we move to a different data store.

app/controllers/resources_controller.rb
jcoyne: Again, this is synched to the parent, since we don't track file versions separately in Fedora. I think this could change when we move to a different data store.

justinlittman (approved):

PR #259, opened 2021-01-05
upgrade to rails 6.1

jmartin-sul:
## Why was this change made?

keep things well maintained

unit tests (will also keep an eye on stage and prod after deployment)

lots of rails configs


config/initializers/filter_parameter_logging.rb
```diff
@@ -3,4 +3,6 @@
 # Be sure to restart your server when you modify this file.
 
 # Configure sensitive parameters which will be filtered from the log file.
-Rails.application.config.filter_parameters += [:password]
+Rails.application.config.filter_parameters += %i[
+  password passw secret token _key crypt salt certificate otp ssn
```
jmartin-sul, line 7: the new default file from rails omitted `password`.  i wasn't sure if that was already a default in `filter_parameters`, so i added it here to be safe, but happy to remove if it's redundant.

config.ru
```diff
@@ -5,3 +5,4 @@
 require_relative 'config/environment'
 
 run Rails.application
```
jcoyne, line 7: I don't think we need both of these lines, do we?

jcoyne (changes requested):

config.ru
jcoyne: Looks like we do in fact need both.

config/application.rb
```diff
@@ -1,7 +1,5 @@
 # frozen_string_literal: true
 
-require_relative 'boot'
```
jcoyne, line 3: I'm pretty sure we need this line: https://github.com/rails/rails/blob/master/railties/lib/rails/generators/rails/app/templates/config/application.rb.tt#L1

jcoyne (changes requested):

config.ru
jmartin-sul: ah yeah, wasn't sure about that.  preference for which to keep?

config/application.rb
jmartin-sul: i was also surprised by the disappearance of that line, i'm pretty sure this came from the `rails app:update` invocation, but also could've been bad merging on my part.  will add back...

jcoyne: It is there in a newly generated rails 6.1, so presumably we need it.

jcoyne (approved):

PR #276, opened 2021-02-27
tests for passing accession flag to /resource, rubocop touchups

jmartin-sul:
pass as URL param, as sdr-client does.

## Why was this change made?

show that the `accession` parameter is passed along correctly to `IngestJob` by the `/resources` route.

see also https://github.com/sul-dlss/sdr-client/pull/148

tests only

n/a

### DIFF
--- a/spec/requests/create_resource_spec.rb
+++ b/spec/requests/create_resource_spec.rb
@@ -156,6 +156,30 @@ RSpec.describe 'Create a resource' do
                                                               start_workflow: nil)
     end
 
+    context 'when the accession flag is explcitly set to true' do
+      it 'kicks off accession workflow' do
+        post '/v1/resources?accession=true',
+             params: request,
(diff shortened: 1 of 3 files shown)


jmartin-sul:
see also see also sul-dlss/happy-heron#1191 and https://github.com/sul-dlss/sdr-client/pull/148

spec/requests/create_resource_spec.rb
```diff
@@ -141,9 +141,10 @@
       file_params['size'] = 10
       model_params
     end
+    let(:path) { '/v1/resources' }
```
jcoyne, line 144: I find that putting this as a variable is harder to read than how it was before.  This isn't used more than one time, so why?

jmartin-sul: was cruft from a slightly different implementation, moved back inline

jmartin-sul:
suggestion taken, threw in a bit of rubocop while i was at it

jcoyne (approved):

PR #285, opened 2021-03-19
modify UpdateJob so it accessions by default

jmartin-sul:
## Why was this change made?

also, add slightly stricter version checking, and alert when there are versioning problems

connects to https://github.com/sul-dlss/happy-heron/issues/1248

- [x] unit tests
  - [x] add a test for versions match but object is open
  - [x] test for HB notifications
- [x] deploy to QA
  - [x] initial PURL res descriptive metadata now fails roundtrip validation when registering the SDR object, so testing of this change (which modifies the update call made after the purl is successfully reserved) is temporarily blocked.  see https://app.honeybadger.io/projects/50568/faults/78470985
    - `Cocina::RoundtripValidationError: Roundtripping of descriptive metadata unsuccessful. Expected {:title=>[{:value=>"purl res, sdr-api commit 3b8a1d6b"}], :contributor=>[], :event=>[], :form=>[{:structuredValue=>[{:value=>"PURL reservation", :type=>"type"}], :type=>"resource type", :source=>{:value=>"Stanford self-deposit resource types"}}], :note=>[{:value=>"", :type=>"summary"}, {}], :subject=>[], :relatedResource=>[]} but received {:title=>[{:value=>"purl res, sdr-api commit 3b8a1d6b"}], :form=>[{:type=>"resource type", :source=>{:value=>"Stanford self-deposit resource types"}, :structuredValue=>[{:value=>"PURL reservation", :type=>"type"}]}]}.`
    - sounds like this is actually a more general H2 mapping problem, see https://github.com/sul-dlss/dor-services-app/issues/2557
  - [x] test on QA

ticketed an integration test:  https://github.com/sul-dlss/infrastructure-integration-test/issues/211

### DIFF
--- a/app/jobs/update_job.rb
+++ b/app/jobs/update_job.rb
@@ -13,7 +13,7 @@ class UpdateJob < ApplicationJob
   # @param [BackgroundJobResult] background_job_result
   # rubocop:disable Metrics/AbcSize
   # rubocop:disable Metrics/MethodLength
-  def perform(model_params:, signed_ids:, background_job_result:)
+  def perform(model_params:, signed_ids:, background_job_result:, start_workflow: true)
     # Increment the try count
     background_job_result.try_count += 1
     background_job_result.processing!
@@ -22,25 +22,35 @@ class UpdateJob < ApplicationJob
 
     object_client = Dor::Services::Client.object(model.externalIdentifier)
     existing = object_client.find
-    if existing.version >= model.version
-      background_job_result.output = {
-        errors: [
-          title: 'Version conflict',
-          detail: "The repository already has a version '#{existing.version}' for " \
-                  "#{existing.externalIdentifier}, and you provided '#{model.version}'"
-        ]
-      }
+
+    unless [existing.version, existing.version + 1].include?(model.version)
+      error_title = 'Version conflict'
+      error_detail = "The repository is on version '#{existing.version}' for #{existing.externalIdentifier}. " \
+                     'You may either: update the current version (for v1 registered, or a later open version); ' \
+                     "or open a new version.  You tried to create/update version '#{model.version}'."
+
+      Honeybadger.notify("#{error_title}: #{error_detail}")
+      background_job_result.output = { errors: [title: error_title, detail: error_detail] }
       background_job_result.complete!
+
       return
     end
 
-    object_client.version.open
     object_client.update(params: model)
 
     background_job_result.output = { druid: model.externalIdentifier }
 
+    versioning_params = { description: 'Update via sdr-api', significance: 'major' }
     StageFiles.stage(signed_ids, model.externalIdentifier) do
-      object_client.version.close(description: 'Update via sdr-api', significance: 'major')
+      if start_workflow
+        # this will check openability, open/close a version as needed, and then kick off accessioning after
+        # that regardless of whether a version was opened/closed.
+        object_client.accession.start(versioning_params.merge(workflow: 'accessionWF'))
+      elsif model.version == existing.version + 1
+        # don't kick off accessioning, just create a new version where only metadata was updated
+        object_client.version.open
+        object_client.version.close(versioning_params.merge(start_accession: false))
+      end
     end
 
     background_job_result.complete!
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -30,6 +30,9 @@ RSpec/ExampleLength:
 RSpec/MultipleExpectations:
   Enabled: false
 
+RSpec/MultipleMemoizedHelpers:
+  Enabled: false
+
 RSpec/NestedGroups:
   Max: 4
 
--- a/spec/jobs/update_job_spec.rb
+++ b/spec/jobs/update_job_spec.rb
@@ -22,7 +22,7 @@ RSpec.describe UpdateJob, type: :job do
     {
       type: Cocina::Models::Vocab.book,
       label: 'hello',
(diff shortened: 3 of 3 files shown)


app/jobs/update_job.rb
```diff
@@ -13,7 +13,8 @@ class UpdateJob < ApplicationJob
   # @param [BackgroundJobResult] background_job_result
   # rubocop:disable Metrics/AbcSize
   # rubocop:disable Metrics/MethodLength
-  def perform(model_params:, signed_ids:, background_job_result:)
+  # rubocop:disable Metrics/PerceivedComplexity
+  def perform(model_params:, signed_ids:, background_job_result:, start_workflow: true)
```
jmartin-sul, line 17: this changes the current default behavior.  wasn't sure which way to go, but since this is consistent with `IngestJob`, and since it's the behavior we seemed to expect by default, i went this way.  happy to change, of course.

app/services/workflow.rb
```diff
@@ -10,6 +10,14 @@ def self.create_unless_exists(druid, workflow_name, version: 1)
     workflow_client.create_workflow_by_name(druid, workflow_name, version: version, lane_id: 'low')
   end
 
+  # Checks if current version has any incomplete wf steps and there is a versionWF
+  # @return [Boolean] true if object is open for versioning
+  def self.open_for_versioning?(druid, version)
```
jmartin-sul, line 15: DSA and argo both do a similar check.  should this be centralized into workflow client?  (not in this ticket necessarily, i'd assume in a follow on if the answer is yes)

app/jobs/update_job.rb
```diff
@@ -13,7 +13,8 @@ class UpdateJob < ApplicationJob
   # @param [BackgroundJobResult] background_job_result
   # rubocop:disable Metrics/AbcSize
   # rubocop:disable Metrics/MethodLength
-  def perform(model_params:, signed_ids:, background_job_result:)
+  # rubocop:disable Metrics/PerceivedComplexity
```
jmartin-sul, line 16: this is a pretty long method with a bunch of branching and multiple ways to return.  but i didn't see any obvious ways to decompose it into something clearer.  the [`if start_workflow ...`](https://github.com/sul-dlss/sdr-api/pull/285/files#diff-2232f4d49d67b6a8ccaec48c4181a71e6e3cb0e28ec7d648aad63c22691dc35bR54-R58) conditional below was the last thing i added, and was what necessitated this exception.

app/jobs/update_job.rb
```diff
@@ -22,25 +23,39 @@ def perform(model_params:, signed_ids:, background_job_result:)
...
 
-    object_client.version.open
     object_client.update(params: model)
 
     background_job_result.output = { druid: model.externalIdentifier }
 
     StageFiles.stage(signed_ids, model.externalIdentifier) do
-      object_client.version.close(description: 'Update via sdr-api', significance: 'major')
+      if start_workflow
+        Workflow.create_unless_exists(model.externalIdentifier, 'accessionWF', version: existing.version)
```
jmartin-sul, line 55: i wasn't 100% sure `existing.version` was right (as opposed to `model.version`), but i think this is consistent with how `IngestJob` makes the call.

jmartin-sul (commented):
leaving in draft because it needs testing on QA or stage before merge, but ready for review/suggestions, i think

app/jobs/update_job.rb
peetucket: And then get rid of the `version.open` above since this will be handled automatically by that `accession.start` call if needed

app/jobs/update_job.rb
```diff
@@ -22,25 +23,39 @@ def perform(model_params:, signed_ids:, background_job_result:)
 
     object_client = Dor::Services::Client.object(model.externalIdentifier)
     existing = object_client.find
-    if existing.version >= model.version
```
peetucket, line 25: `if existing.version >= model.version && Workflow.open_for_versioning?(model.externalIdentifier, existing.version)` ?

app/jobs/update_job.rb
jmartin-sul: thanks for the pairing on this, @peetucket!  dor-services-client/dor-services-app accessioning did the trick.

jmartin-sul: fwiw, i did end up keeping `version.open` for the case where a new version is being updated, but no accessioning has done.

jmartin-sul:
tested on QA.  see:
https://argo-qa.stanford.edu/view/gh811dp6442
https://sul-purl-stage.stanford.edu/gh811dp6442

#### after initial title entry for purl reservation
Screen Shot 2021-03-31 at 11 32 41 PM

#### after picking a type for the purl res (and viewing dashboard from another tab)
Screen Shot 2021-03-31 at 11 33 09 PM

#### fleshing out the purl into a real deposit for accessioning
Screen Shot 2021-03-31 at 11 34 32 PM

#### metadata filled in, file attached, deposit triggered and completed
Screen Shot 2021-03-31 at 11 35 36 PM

jcoyne (approved):

app/jobs/update_job.rb
peetucket: Not sure if it matters, but one thing to note is that we don't kick off accessionWF after opening and closing a version, you won't get a publish.  This means that if the use case for this is a metadata only change, the PURL won't be updated.  My understanding was always that a version with accessioning was needed for any changes that manifest on the PURL for this reason.

PR #303, opened 2021-04-20
Update for changes to access/embargo in cocina models.

justinlittman:
## Why was this change made?
Cocina models now enforces valid access/embargo values.

Unit

NA

### DIFF
--- a/spec/jobs/ingest_job_spec.rb
+++ b/spec/jobs/ingest_job_spec.rb
@@ -24,11 +24,13 @@ RSpec.describe IngestJob, type: :job do
       version: 1,
       access: {
         copyright: 'All rights reserved unless otherwise indicated.',
+        access: 'world',
         download: 'none',
         useAndReproductionStatement: 'Property rights reside with the repository...',
         embargo: {
           releaseDate: '2029-06-22T07:00:00.000+00:00',
           access: 'world',
+          download: 'world',
           useAndReproductionStatement: 'Whatever you want'
         }
       },
@@ -65,7 +67,8 @@ RSpec.describe IngestJob, type: :job do
         shelve: true
       },
       access: {
-        access: 'citation-only'
+        access: 'stanford',
+        download: 'stanford'
       },
       hasMessageDigests: [
         { type: 'md5', digest: '7f99d78a78a233ebbf81ec5b364380fc' },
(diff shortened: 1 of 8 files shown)


spec/jobs/ingest_job_spec.rb
```diff
@@ -65,7 +67,8 @@
         shelve: true
       },
       access: {
-        access: 'citation-only'
+        access: 'stanford',
+        download: 'stanford'
```
ndushay, line 71: would we in any way be better served keeping this citation only or the new equivalent and making a separate test for stanford access?

spec/jobs/update_job_spec.rb
```diff
@@ -67,7 +69,8 @@
         shelve: true
       },
       access: {
-        access: 'citation-only'
+        access: 'stanford',
+        download: 'stanford'
```
ndushay, line 73: same question here

spec/requests/create_resource_spec.rb
```diff
@@ -110,8 +111,8 @@
                       "shelve":true
                     },
                     "access": {
-                      "access":"citation-only",
-                      "download":"none"
+                      "access":"stanford",
+                      "download":"stanford"
```
ndushay, line 115: and here

spec/requests/update_resource_spec.rb
```diff
@@ -66,8 +67,8 @@
                     "shelve":true
                   },
                   "access": {
-                    "access":"citation-only",
-                    "download":"none"
+                    "access":"stanford",
+                    "download":"stanford"
```
ndushay, line 71: and here.

ndushay (commented):
wondering if changing the "citation-only" examples to "stanford" permissions in all places is missing an opportunity to try some diff settings ...

Otherwise looks ok.

spec/jobs/ingest_job_spec.rb
justinlittman: `citation-only` isn't a valid value for a file. The actual values don't matter, as long as they're valid.

spec/jobs/update_job_spec.rb
justinlittman: Same answer as above.

spec/requests/create_resource_spec.rb
justinlittman: Same. `citation-only` isn't valid.

spec/requests/update_resource_spec.rb
justinlittman: Guess what my answer is going to be?

jcoyne (approved):
looks fine to me

PR #325, opened 2021-06-29
UpdateJob - do not retry BadRequest or Conflict errors from DSA

### LINKED ISSUE #324: Don't retry when validation error (409) from dor-services-app

ndushay:
## Why was this change made?

We shouldn't automatically retry jobs when they have errors that won't resolve themselves.

Fixes #324

unit tests.

### DIFF
--- a/app/jobs/update_job.rb
+++ b/app/jobs/update_job.rb
@@ -55,6 +55,17 @@ class UpdateJob < ApplicationJob
     end
 
     background_job_result.complete!
+  rescue Dor::Services::Client::BadRequestError => e
+    # report as error and do not retry
+    error = { errors: [title: 'HTTP 400 (Bad Request) from dor-services-app', message: e.message] }
+    background_job_result.output = background_job_result.output.merge(error)
+    background_job_result.complete!
+  rescue Dor::Services::Client::ConflictResponse => e
+    # RoundtripValidationError for cocina uses HTTP 409 as status
+    # report as error and do not retry
+    error = { errors: [title: 'HTTP 409 (Conflict) from dor-services-app', message: e.message] }
+    background_job_result.output = background_job_result.output.merge(error)
+    background_job_result.complete!
   rescue StandardError => e
     # This causes Sidekiq to retry.
     if background_job_result.try_count < Settings.sdr_api.ingest_retries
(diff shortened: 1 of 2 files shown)


app/jobs/update_job.rb
```diff
@@ -55,6 +55,17 @@ def perform(model_params:, signed_ids:, background_job_result:, start_workflow:
     end
 
     background_job_result.complete!
+  rescue Dor::Services::Client::BadRequestError => e
+    # report as error and do not retry
+    background_job_result.output = { errors: [title: 'HTTP 400 (Bad Request) from dor-services-app',
```
jmartin-sul, line 60: might want to do the `background_job_result.output.merge` like in the `rescue StandardError` case, so as to keep the `{ druid: model.externalIdentifier }` from line 42?  or, will this always come from `object_client.update` on line 40, implying that `background_job_result.output` will never be set when this error is hit?

app/jobs/update_job.rb
```diff
@@ -55,6 +55,17 @@ def perform(model_params:, signed_ids:, background_job_result:, start_workflow:
...
+                                              message: e.message] }
+    background_job_result.complete!
+  rescue Dor::Services::Client::ConflictResponse => e
+    # RoundtripValidationError for cocina uses HTTP 409 as status
+    # report as error and do not retry
+    background_job_result.output = { errors: [title: 'HTTP 409 (Conflict) from dor-services-app',
```
jmartin-sul, line 66: same question here

jmartin-sul (commented):
one small worry, but otherwise LGTM

ndushay:
@jmartin-sul - good catch - fixed.

app/jobs/update_job.rb
ndushay: good point, fixed.

app/jobs/update_job.rb
ndushay: good point, fixed.

jmartin-sul (approved):

jmartin-sul:
> @jmartin-sul - good catch - fixed.

thanks!  i saw that you requested review from justin, so refraining from merging, but feel free to merge if you were just looking for one review to sign off.

PR #333, opened 2021-07-12
Update openapi to allow our testing prefix through

jcoyne:


ndushay (commented):
I would have no idea that is a testing prefix.  Is it documented somewhere?

jcoyne:
@ndushay yes, here:
https://github.com/sul-dlss/cocina-models/blob/bf68de4a620fe16b98f260c4a6e2451ddd228019/openapi.yml#L881-L882

ndushay (approved):

PR #396, opened 2022-02-18
add an okcomputer check for sidekiq process and worker counts

### LINKED ISSUE #381: Add an okcomputer check to confirm that the correct number of workers are running
For apps that use sidekiq, add an okcomputer check to confirm that the correct number of workers are running.  We do a similar thing in some resque-pool using apps (e.g. pres cat).  This is not just a minimum check, but checks that the configured number of workers are present to ensure there are no stale older workers still running unexpectedly.

@jcoyne points us to some useful API documentation:  https://github.com/mperham/sidekiq/wiki/API#workers

This ticket is being added to sdr-api repo, but it could apply to any repo that uses sidekiq.

jmartin-sul:

fixes #381

after rebasing on the latest `main` (to pick up the upgrade to sidekiq 7), i deployed this branch to QA, and tested the following scenarios:

### web only host

N/A

#### normal operation (exactly the number of expected management processes and worker threads)

Screen Shot 2023-03-30 at 8 48 06 PM

#### some but not all worker management processes running

Screen Shot 2023-03-30 at 8 48 38 PM

#### no worker management processes running

Screen Shot 2023-03-30 at 8 49 00 PM

#### more than expected number of worker management processes running

Screen Shot 2023-03-30 at 8 49 49 PM

⚡ ⚠ If this change has cross service impact, ***run [integration tests](https://github.com/sul-dlss/infrastructure-integration-test)*** and/or test in [stage|qa] environment, in addition to specs. ⚡

### DIFF
--- a/config/initializers/okcomputer.rb
+++ b/config/initializers/okcomputer.rb
@@ -31,8 +31,106 @@ class TablesHaveDataCheck < OkComputer::Check
   end
 end
 
+# confirm that the expected number of sidekiq worker processes and threads are running
+class SidekiqWorkerCountCheck < OkComputer::Check
+  class ExpectedEnvVarMissing < StandardError; end
+
+  def check # rubocop:disable Metrics/MethodLength
+    actual_local_sidekiq_processes = fetch_local_sidekiq_processes
+    actual_local_sidekiq_process_count = actual_local_sidekiq_processes.size
+    actual_local_total_concurrency = actual_local_sidekiq_processes.sum { |process| process['concurrency'] }
+
+    error_list = calculate_error_list(actual_local_sidekiq_process_count: actual_local_sidekiq_process_count,
+                                      actual_local_total_concurrency: actual_local_total_concurrency)
+
+    if error_list.empty?
+      mark_message "Sidekiq worker counts as expected on this VM: #{actual_local_sidekiq_process_count} worker " \
+                   "processes, #{actual_local_total_concurrency} concurrent worker threads total."
+    else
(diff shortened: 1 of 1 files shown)


jmartin-sul:
i tried this on stage, and even though i could see the env var in both the shell and from rails console, the app couldn't see it.  maybe it's only available in login shells?  will work w/ ops to debug.

either way, will clean up the lint before taking out of draft, of course.

peetucket:
Todo:  figure out why `ENV['EXPECTED_SIDEKIQ_PROC_COUNT']` is not being picked up by the rails app even though it works on the rails console.

peetucket:
Is it possible the web application is running in the context of a user other than the "sdr"?  It looks like the ENV variable is being set by puppet for the `sdr` user: https://github.com/sul-dlss/puppet/blob/3f91ff944afed6f28cb2a5be819e7ea121eca2d2/hieradata/node/sdr-api-qa-01.stanford.edu.eyaml#L49

This would make it available on the rails console when you ssh in, but not to the rails app (if the rails app is running in the context of a different user, eg `apache`).

jmartin-sul:
well, for whatever reason, the env var worked just fine when i came back to this PR this week 🤷 (most of the work ended up being around writing something that'd work for both envs with a single worker VM and envs with multiple worker VMs, since i wanted to write this with an eye towards re-usability across our various apps)

mjgiarlo (approved):

mjgiarlo:
@jmartin-sul Approved. Shall we ship this or is there anything left to do?

PR #401, opened 2022-02-25
Update cocina model.

justinlittman:
Progress

⚡ ⚠ If this change has cross service impact, ***run [integration tests](https://github.com/sul-dlss/infrastructure-integration-test)*** and/or test in [stage|qa] environment, in addition to specs. ⚡

Unit

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -20,10 +20,10 @@ gem 'bootsnap', '>= 1.4.2', require: false
 
 gem 'action_policy'
 gem 'assembly-objectfile', '~> 1.9'
-gem 'cocina-models', '~> 0.65.0'
+gem 'cocina-models', '~> 0.66.0'
 gem 'committee'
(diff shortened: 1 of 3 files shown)


Gemfile
```diff
@@ -20,10 +20,10 @@ gem 'bootsnap', '>= 1.4.2', require: false
 
 gem 'action_policy'
 gem 'assembly-objectfile', '~> 1.9'
-gem 'cocina-models', '~> 0.65.0'
+gem 'cocina-models', '~> 0.66.0'
 gem 'committee'
 gem 'config', '~> 2.0'
-gem 'dor-services-client', '~> 7.8'
+gem 'dor-services-client', '~> 7.12'
```
jcoyne, line 26: This change isn't *really* necessary, because '~> 7.8' matches with 7.12.0, but NBD also.

jcoyne (approved):

PR #410, opened 2022-03-08
Skip managing files when Cocina file URIs present

mjgiarlo:
Connects to sul-dlss/sdr-client#184

`sdr-api` receives requests from both:

1. systems like H2 that rely on the API to deposit files to SDR; and
1. users hand-creating objects via the sdr-client CLI.

The latter use case allows a user to update an existing SDR object, e.g., to amend an item's APO. This operation does not require sdr-api to handle files and is merely passing through Cocina to SDR. One way we can tell whether a Cocina structure depends on sdr-api to manage files is by sniffing files' external identifiers. If the external identifier of a file is a legitimate HTTP(S) URI, SDR already has a file on hand for the object, and sdr-api can simply pass through the structure undecorated. If on the other hand the external identifier is not an HTTP(S) URI, that is a signal that the originating user or system expects the API to manage files for them.

CI

### DIFF
--- a/app/controllers/resources_controller.rb
+++ b/app/controllers/resources_controller.rb
@@ -99,6 +99,22 @@ class ResourcesController < ApplicationController
     file_sets(model_params).each do |fileset|
       fileset[:version] = model_params[:version]
       fileset.dig(:structural, :contains).each do |file|
+        # NOTE: sdr-api receives requests from both:
+        #
+        #   1. systems like H2 that rely on the API to deposit files to SDR; and
+        #   2. users hand-creating objects via the sdr-client CLI.
+        #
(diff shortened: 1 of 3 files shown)


jcoyne (approved):

mjgiarlo:
@jcoyne @justinlittman Do you think this needs intentional testing in QA/stage (thinking of H2 and GBooks) or should they be unaffected? I was assuming the latter but thought it would be good to ask.

justinlittman:
Deploying to QA and running H2 seems reasonable just to make sure.

mjgiarlo:
@justinlittman Good call. I deployed to QA and ran the H2 integration test successfully.

PR #471, opened 2022-07-15
Add a parameter for providing a priority to the request

jcoyne:
This gets passed on to the workflow as 'lane'.  This will allow us to give h2 requests a default priority and Google Books a low priority.

We recently reduced the number of preservation robots workers so as not to overload Ceph.  This caused a queue to build up and h2 deposits got stuck behind google books deposits, leading to a degraded user experience.

CI

### DIFF
--- a/app/controllers/resources_controller.rb
+++ b/app/controllers/resources_controller.rb
@@ -14,8 +14,7 @@ class ResourcesController < ApplicationController
   # rubocop:disable Metrics/MethodLength
   def create
     begin
-      request_dro = cocina_request_model(params.except(:action, :controller, :resource, :accession,
-                                                       :assign_doi).to_unsafe_h)
+      request_dro = cocina_request_model
     rescue BlobError => e
       # Returning 500 because not clear whose fault it is.
       return render build_error('500', e, 'Error matching uploading files to file parameters.')
@@ -27,7 +26,8 @@ class ResourcesController < ApplicationController
                             signed_ids: signed_ids(params),
                             background_job_result: result,
                             start_workflow: params.fetch(:accession, false),
-                            assign_doi: params.fetch(:assign_doi, false))
+                            assign_doi: params.fetch(:assign_doi, false),
+                            priority: params.fetch(:priority, 'default'))
 
     render json: { jobId: result.id },
            location: result,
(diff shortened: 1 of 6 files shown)


app/controllers/resources_controller.rb
```diff
@@ -70,6 +70,16 @@ def show
...
   private
 
+  CREATE_PARAMS_EXCLUDE_FROM_COCINA = [:action, :controller, :resource, :accession, :priority, :assign_doi]
+
+  def cocina_create_params
+    params.except(*CREATE_PARAMS_EXCLUDE_FROM_COCINA).to_unsafe_h
+  end
+
+  def cocina_update_params
+    params.except(:action, :controller, :resource, :id).to_unsafe_h
```
edsu, line 80: It wasn't clear to me why creates need to ignore more than updates. Is that worth having a note about?

app/jobs/ingest_job.rb
```diff
@@ -10,12 +10,14 @@ class IngestJob < ApplicationJob
...
+  # @param [BackgroundJobResult] background_job_result
   # @param [Boolean] start_workflow if true, start accessionWF
   # @param [Boolean] assign_doi if true, adds DOI to Cocina obj
-  # @param [BackgroundJobResult] background_job_result
+  # @param [String] priority ('default') determines the relative priority used for the workflow.
+  #                                      Value may be 'low' or 'default'
   # rubocop:disable Metrics/AbcSize
   # rubocop:disable Metrics/MethodLength
-  def perform(model_params:, signed_ids:, background_job_result:, start_workflow: true, assign_doi: false)
+  def perform(model_params:, signed_ids:, background_job_result:, start_workflow: true, assign_doi: false, priority: 'default')
```
edsu, line 20: Would it be worthwhile to use constant values for these priorities, or to check that they are correct, to help ensure they are being used consistently?

edsu (approved):

app/jobs/ingest_job.rb
jcoyne: We already are doing that via the openapi.yml.  If you used something other than default or low it would be an invalid request.

app/controllers/resources_controller.rb
jcoyne: I don't know why. This is just how the code was written. I just moved it to separate method which makes it easier to understand.

PR #480, opened 2022-08-15
Adds support for setting version description on update.

justinlittman:
refs https://github.com/sul-dlss/happy-heron/issues/2090

So that it can be passed by H2.

Unit

### DIFF
--- a/openapi.yml
+++ b/openapi.yml
@@ -199,6 +199,10 @@ paths:
           required: true
           schema:
             $ref: '#/components/schemas/Druid'
+        - name: versionDescription
+          in: query          
+          schema:
+            type: string
       requestBody:
         description: The metadata for the object
         required: true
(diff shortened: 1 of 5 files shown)


openapi.yml
```diff
@@ -199,6 +199,10 @@ paths:
           required: true
           schema:
             $ref: '#/components/schemas/Druid'
+        - name: versionDescription
+          in: query
```
jcoyne, line 203: Are we sure the query is the right place for this? Does this limit the length of the description?  Will this prevent us from using the full UTF-8 character set?

justinlittman: The motivation for this is to keep the cocina params separate from other params. Among other reasons, this avoids the risk of a future developer accidentally clobbering the param when updating cocina. There are other parts of the API where query params are used at the same time as body params.

The default length limitation in Apache is 8190 characters. This seems like a reasonable number, but can be increased if necessary. I believe that URL encoding should give us UTF-8.

jcoyne (approved):

PR #487, opened 2022-08-30
Improve message for errors about version mismatch

jcoyne:

The part about v1 registered was very misleading and it didn't say what the allowed values were.

ci


justinlittman (approved):

ndushay (commented):
who sees these error messages?  Should Andrew have a say in the message content?

jcoyne:
Just people who look at honeybadger to decipher errors

mjgiarlo:
@jcoyne @ndushay @justinlittman a friendly amendment: https://github.com/sul-dlss/sdr-api/pull/490/files

PR #495, opened 2022-09-02
Change strategy for determining if a file is activestorage or existing.

### LINKED ISSUE #493: When doing an update ensure files are known
When updating a file ensure every file that is not an upload is already known by DSA.

Download the cocina representation from DSA and ensure that every file referenced in the new request already exists in DSA.

justinlittman:
closes #493

To more accurately determine if file is in activestorage or existing in SDR.

Unit, QA

### DIFF
--- a/app/controllers/resources_controller.rb
+++ b/app/controllers/resources_controller.rb
@@ -178,13 +178,12 @@ class ResourcesController < ApplicationController
   # amend an item's APO. This operation does not require sdr-api to handle files
   # and is merely passing through Cocina to SDR. One way we can tell whether a Cocina
   # structure depends on sdr-api to manage files is by sniffing files' external
-  # identifiers. If the external identifier of a file is a legitimate HTTP(S) URI,
-  # SDR already has a file on hand for the object, and sdr-api can simply pass through
-  # the structure undecorated. If on the other hand the external identifier is not
-  # an HTTP(S) URI, that is a signal that the originating user or system expects
-  # the API to manage files for them.
+  # identifiers. If the external identifier of a file is a legitimate signed id,
+  # the originating user or system expects the API to manage files for them. On the
+  # other hand, it can be assumed that SDR already has a file on hand for the object,
+  # and sdr-api can simply pass through the structure undecorated.
   def signed_id?(file_id)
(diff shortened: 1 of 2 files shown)


app/controllers/resources_controller.rb
```diff
@@ -173,13 +173,12 @@ def signed_ids(model_params)
...
-  # the structure undecorated. If on the other hand the external identifier is not
-  # an HTTP(S) URI, that is a signal that the originating user or system expects
-  # the API to manage files for them.
+  # identifiers. If the external identifier of a file is a legitimate signed id,
+  # the originating user or system expects the API to manage files for them. On the
+  # other hand, it can be assumed that SDR already has a file on hand for the object,
+  # and sdr-api can simply pass through the structure undecorated.
   def signed_id?(file_id)
-    !file_id.match?(%r{^https?://})
+    ActiveStorage.verifier.valid_message?(file_id)
```
jcoyne, line 181: Does this work with an expired file_id?   Won't it say that it's already in the repo and then let this File into the repo, even though we don't want it to?  

I wonder if we shouldn't just remediate the 2048 identifiers that don't conform to this assumption? https://github.com/sul-dlss/dor-services-app/issues/4220#issuecomment-1235922339

jcoyne (changes requested):

app/controllers/resources_controller.rb
justinlittman: All this does is check that this message was created by this instance of active storage.

It doesn't check if expired or the correct purpose. That is still done here, so none of the verification and error handling is changed: https://github.com/sul-dlss/sdr-api/blob/main/app/controllers/resources_controller.rb#L153

justinlittman: https://api.rubyonrails.org/classes/ActiveSupport/MessageVerifier.html#method-i-valid_message-3F

mjgiarlo:
@justinlittman Shall we ship this? Worth deploying and doing a full integration test run w/ this out there first?

justinlittman:
@mjgiarlo I suggest that this is a better strategy. If you agree, then I think it can be merged.

Issue #522, opened 2022-11-08
Implement user-based access controls so we can enable access for external users


andrewjbtw:
In order to enable API access for external (non-SUL) users, the SDR-API should implement an access control scheme. The current API allows anyone with an account and password to modify any druid. Currently, only a small number of SUL users have accounts and the API operates as essentially an internal service.

While the scheme may grow over time in response to user needs and feedback, we could begin with the following set of permissions, modeled after the H2 permissions:

A non-(super)admin user should be able to

- Create new collections
- Create new items (in any collection? Ideally we'd be able to limit access to collections as we currently do in H2)
- Manage (update) only the collections or items that they "own", where "ownership" means
  - they created the collection/item and have retained ownership
  - ownership has been transferred to them (similar to how we can transfer ownership within H2)   

Since the SDR requires an APO we would need to determine how to configure that for non-admin accounts. Ideally, external users wouldn't need to know even that the concept of the APO exists. Hydrus/H2 have always used APOs but those have not been visible to users.

justinlittman:
"Users" for SDR-API are typically service accounts (not people and definitely on tied to sunets). Is this consistent with these requirements?

andrewjbtw:
Two of the known user groups who have been in touch over the past few years about being able to deposit via API:

- a research lab with a lot of data
- a journalism group who would like to integrate depositing with a system where they take in data from outside sources

andrewjbtw:
Whether we do sunet-based authorization or some other strategy seems more like an implementation question.

justinlittman:
It is an implementation question that would be driven by requirements like "Who/what is a user?"

peetucket:
I don't think we want to tie API "users" to SUNets or specific people - the common pattern with APIs is to have an API key that is issued for a specific application, which is in turn managed by a person (who may change over time as the team that manages the application consuming the API has changes in personnel).  This is, for example, how ORCID manages their external -- we register an application, and then a technical contact.  We then get an API key which grants us certain permissions.  In our case, that API key gets the "create" permissions Andrew described above, along with edit/update permissions to anything created by that specific API key.  If needed, you revoke the API key.

andrewjbtw:
That seems reasonable, and to be clear, I wouldn't have expected something other than API keys. My request is for a limit on what someone with those keys can do. I've used a few APIs managed by organizations I don't work for, and I've gone through registration processes for them, and they've all limited what I can do, so this doesn't seem like an unusual request to me.

[...]

PR #524, opened 2022-11-17
Adds support for auth.

justinlittman:
refs #522

Lock it down.

Unit, integration test


justinlittman:
Here's how SDR client handles:
```
% sdr get druid:xj862qt3229 --url https://sdr-api-qa.stanford.edu
I, [2022-11-17T14:58:44.512849 #45132]  INFO -- : Retrieving metadata from: /v1/resources/druid:xj862qt3229
E, [2022-11-17T14:58:44.964623 #45132] ERROR -- : There was an HTTP 401 error making the request: {"error":"Not Authorized"}
/Users/jlittman/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/sdr-client-0.97.0/lib/sdr_client/find.rb:23:in `run': There was an HTTP 401 error making the request: {"error":"Not Authorized"} (SdrClient::Find::Failed)
```

README.md
```diff
@@ -85,6 +79,91 @@ Given that we have a DRO with two Filesets each with a File (image1.png) and (im
...
+                 :id => 1,
+               :name => nil,
+              :email => "leland@stanford.edu",
+   "password_digest" => "[DIGEST]",
+         :created_at => Thu, 17 Nov 2022 17:26:28.927725000 UTC +00:00,
+         :updated_at => Thu, 17 Nov 2022 17:26:28.927725000 UTC +00:00,
+             :active => true,
+        :full_access => true,
+        :collections => []
+}
```
mjgiarlo, line 96: A minor tweak: what would you think about prompting for the password value so we don't have them stored in shell history? Something like https://github.com/sul-dlss/infrastructure-integration-test/blob/main/spec/support/authentication_helpers.rb#L55-L59

app/models/user.rb
```diff
@@ -8,4 +9,12 @@ class User < ApplicationRecord
...
             length: { minimum: 6 },
             if: -> { new_record? || !password.nil? }
+
+  def active?
+    active
+  end
+
+  def full_access?
+    full_access
+  end
```
mjgiarlo, line 19: Are these needed? I thought ActiveRecord would do this automatically since these are boolean fields.

app/policies/resource_policy.rb
```diff
@@ -6,15 +6,22 @@ class ResourcePolicy < ApplicationPolicy
 
   # Any user with an account can create or update resources
   def create?
-    true
+    return false unless user.active?
+    return true if user.full_access?
+
+    # Collection-limited users can only update DROs
+    return false unless record.dro?
```
mjgiarlo, line 13: I'm not sure I get this. We wanted to allow all active users to create collections, right? Should this be `return true if record.collection?`

db/migrate/20221116163116_add_collections_to_users.rb
```diff
@@ -0,0 +1,8 @@
+class AddCollectionsToUsers < ActiveRecord::Migration[7.0]
+  def change
+    add_column :users, :active, :boolean, default: true
+    add_column :users, :full_access, :boolean, default: true
```
mjgiarlo, line 4: Should we default to false or do you think it's fine to keep the defaults such that they extend the current behavior?

mjgiarlo (commented):
Have some questions for you, none too earth-shattering. In addition:

1. Should we add the 401 response statuses to the API spec or is that extra work that is not helpful?
2. Should we add code that adds a collection to a user's collections (when they are active, non-full access)?
3.

README.md
justinlittman: Done.

app/policies/resource_policy.rb
justinlittman: I'd suggest that in the simplest possible use case, a restricted user can only deposit to a single collection, not create collections. That is certainly the case with the one restricted user that we know about, Google Books. Once we understand other use cases, we can extend the authorization implementation.

db/migrate/20221116163116_add_collections_to_users.rb
justinlittman: As of now, creating a full access user is our normal use case so optimized for that.

app/models/user.rb
justinlittman: Ohhh ... new to me Rails magic. Fixed.

justinlittman:
I'm not sure that adding a typical HTTP response (401) without any sort of unique schema adds value to the openapi. However, if you feel strongly, let me know, and I will add.

db/migrate/20221116163116_add_collections_to_users.rb
mjgiarlo: @justinlittman OK, sounds good.

mjgiarlo:
@justinlittman 💬 
> I'm not sure that adding a typical HTTP response (401) without any sort of unique schema adds value to the openapi. However, if you feel strongly, let me know, and I will add.

As long as we don't run afoul of Committee, I think it's fine?

mjgiarlo (approved):

Issue #526, opened 2022-11-23
Download files from Globus to workspace


mjgiarlo:
We will be using Globus as a means to transfer large files to SDR via H2. Instead of needing many components within SDR to be connected to Globus, we'd like to limit Globus awareness to sdr-api. 

# What's needed

- [x] Add `globus_client` as a dependency
- [ ] When creating or updating a resource, accept one or more new globus-related params giving the `globus_client` the information it needs to find the files (will a directory path e.g. `/uploads/username/work123/version1` suffice?)
- [ ] Set the directory containing the files to read-only (blocked by sul-dlss/globus_client#36)
- [ ] Use this information + the necessary Globus configuration† to download the files and stash them in the DOR workspace
- [ ] Test and verify that what comes back works with recent HFS changes if the filenames expose hierarchy

† This would be a globus client ID and a globus client secret---both of which live in vault (e.g., `vault kv get puppet/application/h2/globus/client_id` & `vault kv get puppet/application/h2/globus/client_secret`)---and also a globus endpoint ID (`34ea3e65-6831-479a-8da3-87f118e3fc2b` for QA, with stage and prod coming soon) and a base directory (`/uploads/`)---both of which could live in `shared_configs` for `sdr-api`.

# Analysis

If helpful, feel free to review the [analysis](https://github.com/sul-dlss/dor-services-app/issues/4270)

justinlittman:
One possible approach is: When constructing the cocina structural metadata, H2 provides a namespaced external identifier that indicates the file's location on globus mount. This is similar to how H2 indicates to SDR API that a file should be copied from ActiveStorage.

mjgiarlo:
@aaron-collier Is this still in progress or should it be closed?

mjgiarlo:
@aaron-collier ☝🏻

PR #531, opened 2022-11-30
[DRAFT] Allow file addition from globus

aaron-collier:

Part of #526 

More to come...


app/controllers/resources_controller.rb
```diff
@@ -127,26 +127,43 @@ def file_identifier(object_id, resource_id)
...
   end
-  # rubocop:enable Metrics/AbcSize
+
+  def metadata_for_file(globus_file, file)
+    return unless File.exist?(globus_file)
+
+    file[:size] = File.size(globus_file)
+  end
+
+  def decorate_file(file:, version:, external_id: nil)
```
ndushay, line 145: I'm very late to the party here, but I have no idea what "decorate_file" is doing.  Could a comment be added?

ndushay (commented):
I'm the least knowledgable about how this is supposed to work, so I'm just commenting.  The code looks fine to me; wondering about adding a comment for one method (I know it was already there but ... it's an opportunity).

Gemfile
```diff
@@ -26,6 +26,7 @@ gem 'config', '~> 2.0'
 gem 'dor-services-client', '~> 12.0'
 gem 'dor-workflow-client', '~> 5.0'
 gem 'druid-tools'
+gem 'globus_client'
```
mjgiarlo, line 29: I was wondering about whether we should pin this or not, but I don't think we need to sweat that now.

app/controllers/resources_controller.rb
```diff
@@ -127,26 +127,43 @@ def file_identifier(object_id, resource_id)
...
+  end
 
   # Decorates the provided FileSets with the information we have in the ActiveStorage table.
   # externalIdentifier is also removed from the request.
   def decorate_request_file_sets(model_params)
     file_sets(model_params).each do |fileset|
       fileset[:version] = 1
       fileset.dig(:structural, :contains).each do |file|
-        decorate_file(file: file, version: 1)
+        external_identifier = signed_id?(file[:externalIdentifier]) ? nil : file[:externalIdentifier]
```
mjgiarlo, line 165: I see we check if the external identifer is a signed_id here, and then we check it again immediately after (in the first line of `#decorate_file`). If we're checking again so soon and we still have the file object (and thus access to the external ID), do you feel we need both calls?

app/controllers/resources_controller.rb
```diff
@@ -158,6 +175,10 @@ def blob_for_signed_id(signed_id, filename)
     raise BlobError, "Unable to find upload for #{filename} (#{signed_id})"
   end
 
+  def file_from_globus(signed_id, filename)
```
mjgiarlo, line 178: I tend to think of `signed_id` as an ActiveStorage-ism. Do we have a signed ID here or just a globus URI? I wonder about changing some of the variable names for extra clarity down the road.

app/services/blobs.rb
```diff
@@ -7,10 +7,17 @@ class Blobs
   def self.blobs_for(signed_ids)
     {}.tap do |blob_hash|
       signed_ids.each do |filename, signed_id|
+        next unless signed_id? signed_id
```
mjgiarlo, line 10: Can/should we align this check with the recent change in the strategy used to determine if an ID is a signed ID or not? E.g.:

```suggestion
        next unless ActiveStorage.verifier.valid_message?(file_id)
```

And then we can ditch the new method below.

app/controllers/resources_controller.rb
aaron-collier: Good catch, probably cruft/oversight. I'll take a look. Thanks :-)

app/controllers/resources_controller.rb
aaron-collier: Sure.

app/services/stage_files.rb
```diff
@@ -6,17 +6,34 @@ def self.stage(signed_ids, druid)
     # Skip side effects if no signed IDs provided
     return yield if signed_ids.empty?
 
+    @dir = StagingDirectory.new(druid: druid, staging_location: Settings.staging_location)
```
mjgiarlo, line 9: It could be a follow-up to this PR, but given that 1) the number of methods in this class has grown, and 2) we are holding onto state (`@dir`), I wonder if we should consider refactoring this class so that `StageFiles.stage` remains the entry but the rest of the work is done at the instance level?

app/services/stage_files.rb
```diff
@@ -6,17 +6,34 @@ def self.stage(signed_ids, druid)
...
+
     blobs = Blobs.blobs_for(signed_ids)
```
mjgiarlo, line 11: This is the line that is causing the `Blobs` class to have to be aware of whether it's getting legit signed IDs or not, right? I wonder if `StageFiles` could have the added smarts instead, so that we can better contain the concern about "am I getting signed IDs or other IDs?" rather than having it leak out.

app/services/stage_files.rb
```diff
@@ -6,17 +6,34 @@ def self.stage(signed_ids, druid)
...
+
+  def self.copy_blobs(blobs)
+    copy_files_to_staging(blobs)
     delete_from_active_storage(blobs.values)
   end
+  private_class_method :copy_blobs
+
+  def self.copy_globus(signed_ids)
+    signed_ids.select { |_key, value| value&.match?(%r{^globus://}) }.each do |filename, globus_path|
+      @dir.copy_file(globus_path.gsub('globus://', Settings.globus_location), filename)
```
mjgiarlo, line 28: How does this `gsub` relate to what's on line 179 of the `ResourcesController`?

mjgiarlo:
Note to whoever merges: please squash-merge if there are multiple commits. Thanks!

Gemfile
justinlittman: Is the globus client used?

app/controllers/resources_controller.rb
```diff
@@ -127,26 +127,43 @@ def file_identifier(object_id, resource_id)
...
     declared_md5 = file[:hasMessageDigests].find { |digest| digest.fetch(:type) == 'md5' }.fetch(:digest)
     calculated_md5 = base64_to_hexdigest(blob.checksum)
     raise BlobError, "MD5 mismatch for #{file[:filename]}" if declared_md5 != calculated_md5
   end
-  # rubocop:enable Metrics/AbcSize
+
+  def metadata_for_file(globus_file, file)
+    return unless File.exist?(globus_file)
+
+    file[:size] = File.size(globus_file)
```
justinlittman, line 142: Don't we also need to add message digests and mime type?

Gemfile
aaron-collier: @justinlittman it is not. It was something we discussed keeping here or including when necessary. I'm happy to drop it to avoid any confusion.

app/controllers/resources_controller.rb
aaron-collier: I'm not sure we have that.

app/services/stage_files.rb
aaron-collier: This was less about legit IDs and the order things happened in. It currently lives in `copy_files_to_staging` so the method it was in was mocked, and thus this was skipped, we can't do that with the need for blob vs. globus as far as I can tell.

app/controllers/resources_controller.rb
aaron-collier: Maybe? tests don't fail and I'm pretty sure we only had them if it was ActiveStorage and not a file on disk before. I can work through it.

mjgiarlo: @justinlittman Aaron and I are going to deploy this to QA tomorrow and do some testing to see what's needed. Thanks for the heads-up.

app/controllers/resources_controller.rb
```diff
@@ -142,18 +140,40 @@ def get_fileset_uuid(external_id)
...
+      { type: 'sha1', digest: Digest::SHA1.file(globus_file).hexdigest }
+    ]
+    file[:hasMimeType] = Marcel::MimeType.for Pathname.new(globus_file)
+  end
+
+  def decorate_file(file:, version:, external_id: nil)
+    if signed_id?(file[:externalIdentifier])
+      blob = blob_for_signed_id(file.delete(:externalIdentifier), file[:filename])
+      metadata_for_blob(blob, file)
+    else
```
mjgiarlo, line 167: Should we consider an `elsif` here that checks if the ID looks like a Globus URI? The removal of the check up above means that the use case of "edit a cocina that already has files" will wind up being considered as a globus thing, I think?

app/controllers/resources_controller.rb
```diff
@@ -173,6 +193,10 @@ def blob_for_signed_id(signed_id, filename)
     raise BlobError, "Unable to find upload for #{filename} (#{signed_id})"
   end
 
+  def file_from_globus(globus_id)
+    globus_id.gsub('globus://', Settings.globus_location)
```
mjgiarlo, line 197: If we expect just one instance of `globus://` in the string, we can consider using `String#sub` instead of `String#gsub`. (The latter suggests there may be multiple instances.)

app/controllers/resources_controller.rb
```diff
@@ -205,6 +229,16 @@ def signed_id?(file_id)
...
   end
 
+  def globus_id?(file_id)
+    file_id =~ %r{globus://}
+  end
+
+  def decoratable_id?(file_id)
+    return false unless signed_id?(file_id) || globus_id?(file_id)
+
+    true
```
mjgiarlo, line 239: 🤔 is this functionally equivalent to

```
signed_id?(file_id) || globus_id?(file_id)
```

? If so, I think we can remove the explicit booleans and conditional.

app/controllers/resources_controller.rb
```diff
@@ -205,6 +229,16 @@ def signed_id?(file_id)
     ActiveStorage.verifier.valid_message?(file_id)
   end
 
+  def globus_id?(file_id)
+    file_id =~ %r{globus://}
```
mjgiarlo, line 233: This will produce a match object which I don't think we care about. Would you consider `file_id.start_with?('globus://')`?

Additionally, I wonder if we want to stick `'globus://'` in a constant up above and use the constant throughout the class.

app/services/staging_directory.rb
```diff
@@ -12,6 +12,7 @@ def initialize(druid:, staging_location:)
   def copy_file(source, dest)
     dest_filepath = File.join(content_dir, dest)
     FileUtils.mkdir_p File.dirname(dest_filepath)
+    Honeybadger.context(source: source, dest: dest, dest_filepath: dest_filepath)
```
mjgiarlo, line 15: Cruft?

config/settings.yml
```diff
@@ -12,7 +12,7 @@ workflow:
   url: 'http://localhost:3001'
 
 staging_location: '/dor/workspace'
-
+globus_location: '/globus/sdr_ingest/qa/uploads/'
```
mjgiarlo, line 15: I'd expect this path in our QA shared_configs. Is there a different path we want to default to? Maybe this one is fine? 🤷🏻‍♂️

config/settings/test.yml
```diff
@@ -1 +1,2 @@
 staging_location: 'tmp/assembly'
+globus_location: 'tmp/globus/'
```
mjgiarlo, line 2: Maybe this could be the default value in `config/settings.yml`?

mjgiarlo (approved):

PR #544, opened 2023-01-10
Refactor to separate signed ids from globus ids

### LINKED ISSUE #537: Consider refactoring around signed ID knowledge
The current design has three different classes (`ResourcesController`, `Blobs`, `StageFiles`) that need to be able to tell signed IDs (from ActiveStorage) from Globus URIs and other IDs. We should consider refactoring this so the knowledge lives in fewer places, *e.g.*, within `ResourcesController`, creating a new sibling method to `#signed_ids` called `#globus_ids`, and then passing both of those sets of IDs to the `IngestJob` and `UpdateJob` so the jobs can act on those IDs appropriately without needing to sniff what they've been given. This is but one example of many possible refactorings and is not intended to be prescriptive.

aaron-collier:

Closes #537 

### DIFF
--- a/app/controllers/resources_controller.rb
+++ b/app/controllers/resources_controller.rb
@@ -38,6 +38,7 @@ class ResourcesController < ApplicationController
     result = BackgroundJobResult.create(output: {})
     IngestJob.perform_later(model_params: JSON.parse(request_dro.to_json), # Needs to be sidekiq friendly serialization
                             signed_ids: signed_ids(params),
+                            globus_ids: globus_ids(params),
                             background_job_result: result,
                             start_workflow: params.fetch(:accession, false),
                             assign_doi: params.fetch(:assign_doi, false),
@@ -65,6 +66,7 @@ class ResourcesController < ApplicationController
     result = BackgroundJobResult.create(output: {})
     UpdateJob.perform_later(model_params: JSON.parse(cocina_dro.to_json), # Needs to be sidekiq friendly serialization
                             signed_ids: signed_ids(params),
+                            globus_ids: globus_ids(params),
                             version_description: params[:versionDescription],
                             background_job_result: result)
 
@@ -162,14 +164,18 @@ class ResourcesController < ApplicationController
     file[:hasMimeType] = Marcel::MimeType.for Pathname.new(globus_file)
   end
 
+  # TODO: Clean this method up
   def decorate_file(file:, version:, external_id: nil)
     if signed_id?(file[:externalIdentifier])
       blob = blob_for_signed_id(file.delete(:externalIdentifier), file[:filename])
       metadata_for_blob(blob, file)
-    else
+    elsif globus_id?(file[:externalIdentifier])
       external_id = file[:externalIdentifier]
       globus_file = file_from_globus(file.delete(:externalIdentifier))
-      metadata_for_file(globus_file, file) if globus_id?(external_id)
+      metadata_for_file(globus_file, file) if globus_id?(external_id)    
+    else
+      external_id = file[:externalIdentifier]
+      file.delete(:externalIdentifier)
     end
 
     # Set file params post-processing
@@ -208,7 +214,18 @@ class ResourcesController < ApplicationController
       file_sets(model_params).flat_map do |fileset|
         fileset.dig(:structural, :contains).filter_map do |file|
           # Only include ActiveStorage signed IDs
-          signed_ids[file[:filename]] = file[:externalIdentifier] if decoratable_id?(file[:externalIdentifier])
+          signed_ids[file[:filename]] = file[:externalIdentifier] if signed_id?(file[:externalIdentifier])
+        end
+      end
+    end
+  end
+
+  def globus_ids(model_params)
+    {}.tap do |globus_ids|
+      file_sets(model_params).flat_map do |fileset|
+        fileset.dig(:structural, :contains).filter_map do |file|
+          # Only include ActiveStorage signed IDs
+          globus_ids[file[:filename]] = file[:externalIdentifier] if globus_id?(file[:externalIdentifier])
         end
       end
     end
@@ -233,10 +250,6 @@ class ResourcesController < ApplicationController
 
   def globus_id?(file_id)
     file_id.start_with?(GLOBUS_PREFIX)
-  end
-
-  def decoratable_id?(file_id)
-    signed_id?(file_id) || globus_id?(file_id)
   end
 
   def base64_to_hexdigest(base64)
--- a/app/controllers/resources_controller.rb
+++ b/app/controllers/resources_controller.rb
@@ -37,8 +37,8 @@ class ResourcesController < ApplicationController
 
     result = BackgroundJobResult.create(output: {})
     IngestJob.perform_later(model_params: JSON.parse(request_dro.to_json), # Needs to be sidekiq friendly serialization
-                            signed_ids: signed_ids(params),
-                            globus_ids: globus_ids(params),
+                            signed_ids: signed_ids,
+                            globus_ids: globus_ids,
                             background_job_result: result,
                             start_workflow: params.fetch(:accession, false),
                             assign_doi: params.fetch(:assign_doi, false),
@@ -65,8 +65,8 @@ class ResourcesController < ApplicationController
 
     result = BackgroundJobResult.create(output: {})
     UpdateJob.perform_later(model_params: JSON.parse(cocina_dro.to_json), # Needs to be sidekiq friendly serialization
-                            signed_ids: signed_ids(params),
-                            globus_ids: globus_ids(params),
+                            signed_ids: signed_ids,
+                            globus_ids: globus_ids,
                             version_description: params[:versionDescription],
                             background_job_result: result)
 
@@ -215,9 +215,9 @@ class ResourcesController < ApplicationController
     model_params.fetch(:structural, {}).fetch(:contains, [])
   end
 
-  def signed_ids(model_params)
+  def signed_ids
     {}.tap do |signed_ids|
-      file_sets(model_params).flat_map do |fileset|
+      file_sets(params).flat_map do |fileset|
         fileset.dig(:structural, :contains).filter_map do |file|
           # Only include ActiveStorage signed IDs
           signed_ids[file[:filename]] = file[:externalIdentifier] if signed_id?(file[:externalIdentifier])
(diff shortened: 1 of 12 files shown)


justinlittman (approved):

app/controllers/resources_controller.rb
```diff
@@ -208,7 +220,18 @@ def signed_ids(model_params)
...
+        end
+      end
+    end
+  end
+
+  def globus_ids(model_params)
+    {}.tap do |globus_ids|
+      file_sets(model_params).flat_map do |fileset|
+        fileset.dig(:structural, :contains).filter_map do |file|
+          # Only include ActiveStorage signed IDs
```
lwrubel, line 233: Is this comment still correct?  (Are globus_ids also ActiveStorage signed_ids?)

aaron-collier: Good catch, I changed this but didn't commit it.. whoops... pushed now.

app/jobs/ingest_job.rb
```diff
@@ -48,9 +48,7 @@ def perform(model_params:, signed_ids:, background_job_result:,
     # Create workflow destroys existing steps if called again, so need to check if already created.
     Workflow.create_unless_exists(druid, 'registrationWF', version: 1, priority: priority)
 
-    StageFiles.stage(signed_ids, druid) do
-      Workflow.create_unless_exists(druid, 'accessionWF', version: 1, priority: priority) if start_workflow
-    end
+    stage_files(signed_ids: signed_ids, globus_ids: globus_ids, druid: druid, priority: priority) if start_workflow
```
mjgiarlo, line 51: Looking at this delta, beforehand, we would stage files all the time and create the workflow if `start_workflow` was truthy. Now it looks like we only stage files if `start_workflow` is truthy. Did I get that right? Should that behavior change be in the refactoring? (Or is it harmless?)

mjgiarlo: Related to ☝🏻, I wonder if it's worth inlining the code so that the delta here has:

```ruby
StageBlobs.stage(signed_ids, druid) do
 Workflow.create_unless_exists(druid, 'accessionWF', version: 1, priority: priority) if start_workflow
end

StageGlobus.stage(globus_ids, druid) do
  Workflow.create_unless_exists(druid, 'accessionWF', version: 1, priority: priority) if start_workflow
end
```

❓

app/jobs/update_job.rb
```diff
@@ -48,7 +48,7 @@ def perform(model_params:, signed_ids:, background_job_result:, start_workflow:
 
     versioning_params = { description: version_description || 'Update via sdr-api', significance: 'major' }
 
-    StageFiles.stage(signed_ids, model.externalIdentifier) do
+    StageBlobs.stage(signed_ids, model.externalIdentifier) do
```
mjgiarlo, line 51: Does the `UpdateJob` not need to stage Globus too?

app/services/stage_blobs.rb
```diff
@@ -1,39 +1,24 @@
 # frozen_string_literal: true
 
 # Moves files from the ActiveStorage to the staging mount
-class StageFiles
+class StageBlobs
```
mjgiarlo, line 4: Dig the rename!

app/services/stage_globus.rb
```diff
@@ -0,0 +1,15 @@
...
+# Moves files from the ActiveStorage to the staging mount
+class StageGlobus
+  def self.stage(signed_ids, druid)
+    # Skip side effects if no signed IDs provided
+    return yield if signed_ids.empty?
+
+    dir = StagingDirectory.new(druid: druid, staging_location: Settings.staging_location)
+    signed_ids.select { |_key, value| value&.match?(%r{^globus://}) }.each do |filename, globus_path|
+      dir.copy_file(globus_path.gsub('globus://', Settings.globus_location), filename)
+    end
```
mjgiarlo, line 12: I assume all the `signed_ids` here are Globus IDs. Worth renaming to avoid confusion?

spec/jobs/ingest_job_spec.rb
```diff
@@ -128,7 +129,8 @@
     before do
       described_class.perform_now(model_params: model,
                                   background_job_result: result,
-                                  signed_ids: signed_ids,
+                                  signed_ids: {},
+                                  globus_ids: signed_ids,
```
mjgiarlo, line 133: It looks like this test is using `signed_ids` for both ActiveStorage file IDs and Globus file IDs. I wonder if it's worth renaming that `let` variable to something like `file_ids`?

app/services/stage_globus.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+# Moves files from the ActiveStorage to the staging mount
+class StageGlobus
```
mjgiarlo, line 4: Worth adding unit tests for `StageGlobus`? Maybe you can pull one or more out of the `StageBlobs` spec?

app/jobs/ingest_job.rb
aaron-collier: Works for me :-) I'll update.

aaron-collier: @mjgiarlo actually - I the problem I ran into with that is that we'll always create 2 workflows now... so I wonder if it would be better to add a `Workflow.create...` if there are no signed_ids or globus_ids. Thoughts?

mjgiarlo: Hmmm, shouldn't this logic take care of not creating a duplicate workflow? https://github.com/sul-dlss/sdr-api/blob/main/app/services/workflow.rb#L11

spec/requests/create_dro_spec.rb
```diff
@@ -168,7 +173,7 @@
 
   context 'when the signed_id indicates a globus file' do
     let(:signed_id) { 'globus://abc123/file2.txt' }
```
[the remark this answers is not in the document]
aaron-collier: This one is actually necessary as it's used in the request above. I'm leaning towards a comment on here to avoid confusion rather than duplicating the request setup from above.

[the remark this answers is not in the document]
mjgiarlo: 🆗

app/jobs/update_job.rb
```diff
@@ -15,7 +17,14 @@ class UpdateJob < ApplicationJob
   # @param [String] version_description
   # rubocop:disable Metrics/AbcSize
   # rubocop:disable Metrics/MethodLength
-  def perform(model_params:, signed_ids:, background_job_result:, start_workflow: true, version_description: nil)
+  def perform(**kwargs)
```
mjgiarlo, line 20: Can you say more about changing `UpdateJob#perform` from keyword arguments to a hash? Is it Rubocop-related? I'm curious because `IngestJob#perform` is still using keyword arguments and these jobs are siblings of sorts.

app/services/stage_blobs.rb
```diff
@@ -1,39 +1,24 @@
...
-      copy_blobs(blobs)
-    end
-    yield
-  end
+    yield if blobs.empty?
 
-  def self.copy_blobs(blobs)
-    copy_files_to_staging(blobs)
+    copy_files_to_staging(druid, blobs)
+    yield
```
mjgiarlo, line 13: There are now a few potential `yield`s in close proximity to one another (lines 7, 10, 13). Do we think only one of these will be triggered or is it intentional to have this method `yield` 2-3 times?

app/services/stage_globus.rb
```diff
@@ -0,0 +1,15 @@
...
+  def self.stage(globus_ids, druid)
+    # Skip side effects if no signed IDs provided
+    return yield if globus_ids.empty?
```
mjgiarlo, line 7: I'm wondering whether we would expect this method to yield if in fact it's not doing anything else. For instance, with:

```ruby
[].each do |member|
  Workflow.do_a_thing!
end
```

Would we expect `Workflow.do_a_thing!` to be executed? I don't *think* I would? 🤔

spec/requests/create_dro_spec.rb
```diff
@@ -105,6 +106,7 @@
       expect(IngestJob).to have_received(:perform_later).with(model_params: expected_model_params,
                                                               background_job_result: instance_of(BackgroundJobResult),
                                                               signed_ids: { 'file2.txt' => signed_id },
+                                                              globus_ids: {},
```
mjgiarlo, line 109: Seeing that you had to add this in a few different spots, I wonder if these jobs ought to default to `globus_ids: {}` in the `#perform` method definition to cut down on needing all this setup?

spec/requests/create_dro_spec.rb
```diff
@@ -196,7 +201,8 @@
       expect(response).to have_http_status(:created)
       expect(IngestJob).to have_received(:perform_later).with(model_params: expected_model_params,
                                                               background_job_result: instance_of(BackgroundJobResult),
-                                                              signed_ids: signed_ids,
+                                                              signed_ids: {},
```
mjgiarlo, line 204: Wondering the same about `signed_ids: {}` defaulting in the job's `#perform` method.

app/controllers/resources_controller.rb
```diff
@@ -65,6 +66,7 @@ def update
     result = BackgroundJobResult.create(output: {})
     UpdateJob.perform_later(model_params: JSON.parse(cocina_dro.to_json), # Needs to be sidekiq friendly serialization
                             signed_ids: signed_ids(params),
+                            globus_ids: globus_ids(params),
```
mjgiarlo, line 69: Is this the automagically provided `params` variable? If so, I don't think it needs to be passed to these methods as it should already be available. (I should have caught this in an earlier review, my bad!)

app/jobs/update_job.rb
aaron-collier: Yes, this was to appease rubocop. I'm OK going back and putting an inline exception, but thought this was more idiomatic.

app/services/stage_blobs.rb
aaron-collier: Good catch, fixed.

app/jobs/update_job.rb
mjgiarlo: The fact that so many methods in sdr-api use more than 4-5 keyword args is definitely a smell, IMO! I think in this case, I may be slightly biased in favor of internal consistency within the codebase until we find abstractions to cut down on passing around so much knowledge.

mjgiarlo (approved):

PR #554, opened 2023-02-14
Update cocina-models to 0.87.0 (for Folio record ID support)

mjgiarlo:

To keep pace with cocina-models changes in support of the Folio migration.

CI + will be tested in QA

### DIFF
--- a/openapi.yml
+++ b/openapi.yml
@@ -339,19 +339,6 @@ components:
       required:
         - identifier
         - type
-    Administrative:
-      type: object
-      additionalProperties: false
-      properties:
-        hasAdminPolicy:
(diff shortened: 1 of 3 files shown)


openapi.yml
```diff
@@ -339,19 +339,6 @@ components:
       required:
         - identifier
         - type
-    Administrative:
```
mjgiarlo, line 342: Changes here look funky because I fixed the alpha sort in cocina-models openapi.yml. We're feeling the pain with this bump so that we never have to feel it again.

aaron-collier (approved):

ndushay (commented):
Thanks, Mike

PR #556, opened 2023-02-16
Async digest generation for Globus deposits

edsu:

`GlobusDigestGenerator` will generate digests for Globus deposits if the files currently lack them. It is used during the asynchronous Sidekiq jobs (`UpdateJob` and `IngestJob`) instead of during HTTP response generation when a Resource is created or updated. This means that large files won't cause the HTTP responses to take a long time and risk timeout.

See https://github.com/sul-dlss/happy-heron/issues/2995 for additional context about why this is needed.

### DIFF
--- a/app/controllers/resources_controller.rb
+++ b/app/controllers/resources_controller.rb
@@ -160,11 +160,11 @@ class ResourcesController < ApplicationController
     raise GlobusNotFoundError, "Globus file [#{globus_file}] not found." unless File.exist?(globus_file)
 
     file[:size] = File.size(globus_file)
-    file[:hasMessageDigests] = [
-      { type: 'md5', digest: Digest::MD5.file(globus_file).hexdigest },
-      { type: 'sha1', digest: Digest::SHA1.file(globus_file).hexdigest }
-    ]
     file[:hasMimeType] = Marcel::MimeType.for Pathname.new(globus_file)
+    # file[:hasMessageDigests] are not calculated here since they could take
+    # some time to generate when processing deposits with large files. Instead
+    # digest generation happens in the asynchronous IngestJob or UpdateJob to
+    # avoid a long HTTP response.
   end
 
   def decorate_blob(file)
--- a/app/jobs/ingest_job.rb
+++ b/app/jobs/ingest_job.rb
@@ -24,6 +24,7 @@ class IngestJob < ApplicationJob
     background_job_result.try_count += 1
     background_job_result.processing!
     model = Cocina::Models.build_request(model_params.with_indifferent_access)
+    model = GlobusDigestGenerator.generate(cocina: model, globus_ids: globus_ids)
     begin
       response_cocina_obj = Dor::Services::Client.objects.register(params: model, assign_doi: assign_doi)
       druid = response_cocina_obj.externalIdentifier
--- a/app/jobs/update_job.rb
+++ b/app/jobs/update_job.rb
@@ -51,7 +51,10 @@ class UpdateJob < ApplicationJob
       return
     end
 
-    # Note that not using a lock here since all model params are being provided rather than updating retrieved params.
+    # globus deposits may not have digests yet and they need to be generated before staging (copy)
+    model = GlobusDigestGenerator.generate(cocina: model, globus_ids: globus_ids)
+
+    # not using a lock here since all model params are being provided rather than updating retrieved params.
     object_client.update(params: model, skip_lock: true)
 
     background_job_result.output = { druid: model.externalIdentifier }
@@ -87,7 +90,7 @@ class UpdateJob < ApplicationJob
                                                                                  message: e.message] })
     background_job_result.complete!
   end
-  # rubocop:enable  Metrics/ParameterLists
+  # rubocop:enable Metrics/ParameterLists
   # rubocop:enable Metrics/AbcSize
   # rubocop:enable Metrics/MethodLength
 
--- a/app/services/globus_digest_generator.rb
+++ b/app/services/globus_digest_generator.rb
@@ -1,0 +1,65 @@
+# frozen_string_literal: true
+
+# GlobusDigestGenerator
+#
+# Add digests to Globus DROs if the files currently lack them. Existing files
+# will not be overwritten.
+class GlobusDigestGenerator
+  # @param [Cocina::Model] cocina a DRO to add digests to
+  # @param [Hash] globus_ids a mapping of filenames to their location on disk
+  # @return [Cocina::Model] a new Cocina object with (potentially) new digests
+  def self.generate(cocina:, globus_ids:)
+    new(cocina: cocina, globus_ids: globus_ids).generate
+  end
+
+  def initialize(cocina:, globus_ids:)
+    @cocina = cocina
+    @globus_ids = globus_ids
+  end
+
+  def generate
+    return cocina if !cocina.dro? || globus_ids.blank?
+
+    props = cocina.to_h
+    props[:structural] = generate_filesets(props[:structural])
+
+    if cocina.is_a? Cocina::Models::RequestDRO
+      Cocina::Models.build_request(props)
+    else
+      Cocina::Models.build(props)
+    end
+  end
+
+  private
+
+  attr_reader :cocina, :globus_ids
+
+  def generate_filesets(filesets)
+    filesets[:contains] = filesets[:contains].map do |fileset|
+      generate_files(fileset)
+    end
+
+    filesets
+  end
+
+  def generate_files(fileset)
+    fileset[:structural][:contains] = fileset[:structural][:contains].map do |file|
+      generate_digests(file)
+    end
+
+    fileset
+  end
+
+  def generate_digests(file)
+    # only generate digests for a file if it doesn't have any
+    if file[:hasMessageDigests].blank? && globus_ids.key?(file[:filename])
+      file_path = globus_ids[file[:filename]].sub(%r{^globus://}, Settings.globus_location)
+      file[:hasMessageDigests] = [
+        { type: 'sha1', digest: Digest::SHA1.file(file_path).hexdigest },
+        { type: 'md5', digest: Digest::MD5.file(file_path).hexdigest }
+      ]
+    end
+
+    file
+  end
+end
--- a/app/services/stage_globus.rb
+++ b/app/services/stage_globus.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-# Moves files from the ActiveStorage to the staging mount
+# Moves files from the Globus storage to the staging mount
 # @param [Hash] globus_ids a mapping of filenames to their location on disk
 # @param [String] druid
 # @return [Integer] the number of files staged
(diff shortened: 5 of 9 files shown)


app/jobs/update_job.rb
```diff
@@ -50,7 +50,10 @@ def perform(model_params:,
       return
     end
 
-    # Note that not using a lock here since all model params are being provided rather than updating retrieved params.
+    # globus deposits may not have digests yet and they need to be generated before staging (copy)
+    model = GlobusDigestGenerator.generate(cocina: model, globus_ids: globus_ids)
+
+    # not using a lock here since all model params are being provided rather than updating retrieved params.
     object_client.update(params: model, skip_lock: true)
```
edsu, line 57: I wasn't sure if `skip_lock` should be set to `false` now that it's possible the model params will have been updated?

app/services/globus_digest_generator.rb
```diff
@@ -0,0 +1,67 @@
...
+  def initialize(cocina:, globus_ids:, validate:)
+    @cocina = cocina
+    @globus_ids = globus_ids
+    @validate = validate
+  end
+
+  def generate
+    return cocina unless cocina.dro?
+
+    props = cocina.to_h
```
edsu, line 25: I tried using `new` here instead but because it needs to work with a Dro used for creating or updating a Dro I ran into some trouble with it validating or not. So instead I pressed the easy button and converted the whole thing to a Hash. If there is a better way to do this let me know!

app/services/stage_globus.rb
```diff
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-# Moves files from the ActiveStorage to the staging mount
+# Moves files from the Globus storage to the staging mount
```
mjgiarlo, line 3: Good eye.

app/controllers/resources_controller.rb
```diff
@@ -160,10 +160,7 @@ def metadata_for_file(globus_file, file)
     raise GlobusNotFoundError, "Globus file [#{globus_file}] not found." unless File.exist?(globus_file)
 
     file[:size] = File.size(globus_file)
-    file[:hasMessageDigests] = [
-      { type: 'md5', digest: Digest::MD5.file(globus_file).hexdigest },
-      { type: 'sha1', digest: Digest::SHA1.file(globus_file).hexdigest }
-    ]
+    file[:hasMessageDigests] = [] # calculated during UpdateJob
```
mjgiarlo, line 163: And this needs to be in the controller so that `file[:hasMessageDigests]` is empty, which https://github.com/sul-dlss/sdr-api/pull/556/files#diff-a6913203d143db3d0665441b9a70307c6ee42012b9bc68d5508f2fa3068b338cR57 checks, right? If so, I wonder if tossing that into the comment might help. I can imagine someone (*cough* probably me) looking at this code in the future and thinking "oh, I bet I can delete this..."

app/jobs/ingest_job.rb
```diff
@@ -24,6 +24,7 @@ def perform(model_params:, background_job_result:, signed_ids: {}, globus_ids: {
     background_job_result.try_count += 1
     background_job_result.processing!
     model = Cocina::Models.build_request(model_params.with_indifferent_access)
+    model = GlobusDigestGenerator.generate(cocina: model, globus_ids: globus_ids, validate: false)
```
mjgiarlo, line 27: I'm wondering whose responsibility it ought to be to sniff `globus_ids` to determine if Globus digests need to be calculated. Right now, if I read this right, it seems to be the responsibility of `GlobusDigestGenerator#gen_digests`. I do like the idea of the generator owning this responsibility over sniffing all over the place (in this case, only `UpdateJob` and `IngestJob` for now). Could/should the sniff take places sooner in the generator and result in returning the model back more immediately? I could go either way.

app/jobs/update_job.rb
mjgiarlo: I'm not sure. Maybe @justinlittman has a sense?

app/services/globus_digest_generator.rb
```diff
@@ -0,0 +1,67 @@
+# frozen_string_literal: true
+
+# GlobusDigestGenerator
+#
+# Add digests to Globus DROs if the files currently lack them. Existing files
+# will not be overwritten.
+class GlobusDigestGenerator
+  # @param [Cocina::Model] cocina a DRO to add digests to
+  # @param [Hash] globus_ids a mapping of filenames to their location on disk
+  # @param [Boolean] validate whether to build the new DRO using validation
```
mjgiarlo, line 10: Based on how I'm reading, the `validate` flag controls whether we're creating a `DRO` or a `RequestDRO` (basically the same thing except the latter lacks a druid/external ID). If that's correct, we could consider renaming this argument to something like `request`? Alternatively (again, if this is correct), we could also do away with the flag entirely and let the digest generator sniff the given Cocina for presence/absence of a druid/external ID and build the appropriate object based on that?

app/services/globus_digest_generator.rb
```diff
@@ -0,0 +1,67 @@
...
+
+  attr_reader :cocina, :globus_ids, :validate
+
+  def gen_filesets(filesets)
+    filesets[:contains] = filesets[:contains].map do |fileset|
+      gen_files(fileset)
+    end
+
+    filesets
+  end
```
mjgiarlo, line 45: If the intent here is to take a structural (an array of file sets), map a function over every file set, and then return the result, we could also consider something like:

```suggestion
  def gen_filesets(structural)
    structural.tap do |file_sets|
      file_sets[:contains].map!(&method(:gen_files))
    end
  end
```

Less readable? 🤔

mjgiarlo (commented):
I have a few minor design questions but this is looking good and should help with timeouts and slow requests.

app/jobs/update_job.rb
justinlittman: `skip_lock` should be false since this is always a regeneration of the cocina (even if part of it is generated by SDR API).

`skip_lock` only makes sense when the cocina is being retrieved, modified, and then replaced.

app/services/globus_digest_generator.rb
```diff
@@ -0,0 +1,67 @@
...
+    else
+      Cocina::Models.build_request(props)
+    end
+  end
+
+  private
+
+  attr_reader :cocina, :globus_ids, :validate
+
+  def gen_filesets(filesets)
```
justinlittman, line 39: We typically avoid abbreviations like "gen". (But definitely not a hard rule or anything.)

app/jobs/update_job.rb
edsu: @justinlittman it might not have been clear from the snippet but I believe the Cocina is being fetched on line 34 and then the call to `GlobusDigestGenerator.generate` is adding new digests to the object (if they are missing), and then the Cocina is being saved again?

app/services/globus_digest_generator.rb
edsu: @mjgiarlo yes your interpretation was correct. I thought about having it sniff the Cocina to see if a request was needed, but then I double guessed myself and thought it seemed a bit too magical. But if you had the same thought I'll go with that.

edsu: I tried using `cocina.externalIdentifier` to sniff it but it threw a missing method error when I used it on a RequestDRO. I decided instead I tested to see what type of Cocina it is working with.

app/services/globus_digest_generator.rb
edsu: I actually started with a `tap` but then couldn't quite get `generate_filesets` to work using a similar approach. I ended up thinking there was value in keeping the different generate methods structured similarly. I'd be happy to revisit though.

app/jobs/ingest_job.rb
edsu: Yeah, I wanted to avoid duplicate logic in the various places that call it. You're right though, `generate` could return immediately if `globus_ids` is empty.

app/services/globus_digest_generator.rb
edsu: Makes sense -- I replaced with `generate`.

app/controllers/resources_controller.rb
edsu: @mjgiarlo thanks for noticing this, I could totally see wanting to remove that empty list, and it would break the digest generation. I decided to use `blank?` instead to check if `file[:hasMessageDigests]` has anything in it. This means that supplying an empty list is no longer required.

app/jobs/update_job.rb
justinlittman: The line numbering is off, so I'm going to snippet below and add annotations about what I think is happening:
```
# Model is being built from scratch from provided params.
model = Cocina::Models.build(model_params.with_indifferent_access)

# Existing is being retrieved, but only used to compare versions.
object_client = Dor::Services::Client.object(model.externalIdentifier)
existing = object_client.find

.....
# The built model is being updated.
model = GlobusDigestGenerator.generate(cocina: model, globus_ids: globus_ids)
```
If my reading is correct, then I still think that we want `skip_lock`.

edsu: After chatting about this with @justinlittman we are continuing to go with `skip_lock: true` as it stands, since `model` comes from H2 and cannot be used for locking.

app/controllers/resources_controller.rb
```diff
@@ -160,11 +160,11 @@ def metadata_for_file(globus_file, file)
...
     file[:size] = File.size(globus_file)
-    file[:hasMessageDigests] = [
-      { type: 'md5', digest: Digest::MD5.file(globus_file).hexdigest },
-      { type: 'sha1', digest: Digest::SHA1.file(globus_file).hexdigest }
-    ]
     file[:hasMimeType] = Marcel::MimeType.for Pathname.new(globus_file)
+    # file[:hasMessageDigests] are not calculated here since they could take
+    # some time to generate when processing deposits with large files. Instead
+    # digest generation happens in the asynchronous IngestJob or UpdateJob to
+    # avoid a long HTTP response.
```
mjgiarlo, line 167: 👏🏻

app/services/globus_digest_generator.rb
```diff
@@ -0,0 +1,65 @@
...
+    new(cocina: cocina, globus_ids: globus_ids).generate
+  end
+
+  def initialize(cocina:, globus_ids:)
+    @cocina = cocina
+    @globus_ids = globus_ids
+  end
+
+  def generate
+    return cocina if !cocina.dro? || globus_ids.blank?
```
mjgiarlo, line 21: 😎

mjgiarlo (approved):

app/services/globus_digest_generator.rb
mjgiarlo: OK, np.

app/controllers/resources_controller.rb
mjgiarlo: Perfect 😃

mjgiarlo:
@edsu Feel free to merge when you think the HOLD conditions no longer apply. :shipit:

PR #558, opened 2023-02-22
Ensure AccessionWF is kicked off once

edsu:

We had a situation where large file deposits were triggering a race
condition where multiple accession workflows resulted in techmd getting
confused about missing files.

See https://github.com/sul-dlss/happy-heron/issues/3007

This change ensures that only one AccessionWF is started when staging
from ActiveStorage or Globus. The code assumes that a deposit
stages from ActiveStorage or Globus, but not both.

Integration tests in stage.

### DIFF
--- a/app/jobs/ingest_job.rb
+++ b/app/jobs/ingest_job.rb
@@ -48,13 +48,10 @@ class IngestJob < ApplicationJob
     # Create workflow destroys existing steps if called again, so need to check if already created.
     Workflow.create_unless_exists(druid, 'registrationWF', version: 1, priority: priority)
 
-    StageBlobs.stage(signed_ids, druid) do
-      Workflow.create_unless_exists(druid, 'accessionWF', version: 1, priority: priority) if start_workflow
-    end
+    StageBlobs.stage(signed_ids, druid)
+    StageGlobus.stage(globus_ids, druid)
 
-    StageGlobus.stage(globus_ids, druid) do
-      Workflow.create_unless_exists(druid, 'accessionWF', version: 1, priority: priority) if start_workflow
-    end
+    Workflow.create_unless_exists(druid, 'accessionWF', version: 1, priority: priority) if start_workflow
 
     background_job_result.complete!
   rescue StandardError => e
--- a/app/jobs/update_job.rb
+++ b/app/jobs/update_job.rb
@@ -12,6 +12,7 @@ class UpdateJob < ApplicationJob
 
   # @param [Hash] model_params
   # @param [Hash] filename, signed_ids for the blobs
+  # @param [Hash] filename, globus_ids for the staged Globus files
   # @param [BackgroundJobResult] background_job_result
   # @param [Boolean] start_workflow starts accessionWF if true; if false, opens/closes new version without accessioning
   # @param [String] version_description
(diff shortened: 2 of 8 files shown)


app/jobs/ingest_job.rb
```diff
@@ -48,12 +48,16 @@ def perform(model_params:, background_job_result:, signed_ids: {}, globus_ids: {
...
+    if signed_ids.any?
+      StageBlobs.stage(signed_ids, druid) do
+        Workflow.create_unless_exists(druid, 'accessionWF', version: 1, priority: priority) if start_workflow
+      end
+    elsif globus_ids.any?
+      StageGlobus.stage(globus_ids, druid) do
+        Workflow.create_unless_exists(druid, 'accessionWF', version: 1, priority: priority) if start_workflow
+      end
+    elsif start_workflow
+      Workflow.create_unless_exists(druid, 'accessionWF', version: 1, priority: priority)
```
edsu, line 60: I wasn't sure if this was needed. Do we ever want to kick off accessioning when there are no files to stage?

app/jobs/update_job.rb
```diff
@@ -57,11 +58,15 @@ def perform(model_params:,
...
+    if signed_ids.any?
+      StageBlobs.stage(signed_ids, model.externalIdentifier) do
+        version_or_accession(object_client, model, existing, versioning_params)
+      end
+    elsif globus_ids.any?
+      StageGlobus.stage(globus_ids, model.externalIdentifier) do
+        version_or_accession(object_client, model, existing, versioning_params)
+      end
+    else
       version_or_accession(object_client, model, existing, versioning_params)
```
edsu, line 70: It seems like this is needed when updating metadata only?

app/jobs/ingest_job.rb
```diff
@@ -48,12 +48,16 @@ def perform(model_params:, background_job_result:, signed_ids: {}, globus_ids: {
...
-      Workflow.create_unless_exists(druid, 'accessionWF', version: 1, priority: priority) if start_workflow
-    end
-
-    StageGlobus.stage(globus_ids, druid) do
-      Workflow.create_unless_exists(druid, 'accessionWF', version: 1, priority: priority) if start_workflow
+    if signed_ids.any?
+      StageBlobs.stage(signed_ids, druid) do
+        Workflow.create_unless_exists(druid, 'accessionWF', version: 1, priority: priority) if start_workflow
+      end
+    elsif globus_ids.any?
```
justinlittman, line 55: It should be possible to have both signed_ids and globus_ids.

justinlittman: I think you probably want something like:
```
StageBlobs.stage(signed_ids, druid)
StageGlobus.stage(globus_ids, druid)
Workflow.create_unless_exists(druid, 'accessionWF', version: 1, priority: priority) if start_workflow && (signed_ids.any? || globus_ids.any?)
```

justinlittman: I can't really figure out the purpose of the block in StageBlobs and StageGlobus. I'd recommend getting rid of the block support in these.

edsu: @justinlittman I thought the same at first but then noticed that the AccessionWF would only be kicked off if the Staging succeeded?

edsu: Ok, I've removed the block from both StageGlobus and StageBlobs and updated the tests. It does simplify the way the jobs read.

justinlittman (approved):

PR #568, opened 2023-03-29
Upgrade to Sidekiq7

justinlittman:
Keeping up with the other apps.


Gemfile.lock
```diff
@@ -307,7 +307,8 @@ GEM
       zeitwerk (~> 2.5)
     rainbow (3.1.1)
     rake (13.0.6)
-    redis (4.8.1)
+    redis-client (0.14.0)
```
edsu, line 310: redis-client works the same as redis?

edsu (approved):

PR #585, opened 2023-06-07
[WIP] Override active storage validation of file contents

peetucket:

Attempt to fix sul-dlss/happy-heron#3075


app/controllers/resources_controller.rb
```diff
@@ -150,7 +150,15 @@ def get_fileset_uuid(external_id)
   def metadata_for_blob(blob, file)
     file.delete(:externalIdentifier)
     file[:size] = blob.byte_size
-    file[:hasMimeType] = blob.content_type || 'application/octet-stream'
+    # Invalid JSON files uploaded through H2 will trigger 400 errors in sdr-api since they are parsed
+    # and rejected.  The work around is to change the content_type in H2 for any JSON files to something
+    # specific that will be changed back to application/json after upload is complete.
+    # There is a similar change in H2.  See https://github.com/sul-dlss/happy-heron/issues/3075
```
jmartin-sul, line 156:
```suggestion
    # Invalid JSON files uploaded for deposit with a JSON content type will trigger 400 errors in sdr-api since they are
    # parsed as JSON and rejected.  The work around is to change the content_type in the request for uploads like this to something
    # specific that will be changed back to application/json after upload is complete.
    # There is a corresponding translation in sdr-client.  See https://github.com/sul-dlss/happy-heron/issues/3075
```

for your consideration, a slight tweak to the wording for more specificity (and to update it now that the content-type translation is being done in sdr-client)

spec/requests/create_dro_spec.rb
```diff
@@ -218,6 +219,56 @@
...
+      ActiveStorage::Blob.create!(key: 'tozuehlw6e8du20vn1xfzmiifyok',
+                                  filename: 'file2.txt', content_type: nil,
+                                  byte_size: 10, checksum:)
+    end
+    let(:expected_content_type) { 'application/octet-stream' }
+
+    it 'switches the content type to application/octet-stream' do
+      post '/v1/resources',
+           params: request,
+           headers: { 'Content-Type' => 'application/json', 'Authorization' => "Bearer #{jwt}" }
```
jmartin-sul, line 258: would it be more like actual usage to omit the `Content-Type` header here?

spec/requests/create_dro_spec.rb
```diff
@@ -218,6 +219,56 @@
...
+      ActiveStorage::Blob.create!(key: 'tozuehlw6e8du20vn1xfzmiifyok',
+                                  filename: 'file2.txt', content_type: 'application/x-stanford-json',
+                                  byte_size: 10, checksum:)
+    end
+    let(:expected_content_type) { 'application/json' }
+
+    it 'switches the content type to application/json' do
+      post '/v1/resources',
+           params: request,
+           headers: { 'Content-Type' => 'application/json', 'Authorization' => "Bearer #{jwt}" }
```
jmartin-sul, line 233: not sure either way, but would it be more like actual usage to have the `Content-Type` header be our `application/x-stanford-json` custom type?  or will sdr-client send across `application/json` for this request?

jmartin-sul (approved):
one nitpicky comment suggestion, and two nitpicky questions re: tests, but the application logic change LGTM.

spec/requests/create_dro_spec.rb
peetucket: I believe this `application/json` stays as is since that is the format of wrapper data being posted, and the `request` blob posted is the one with `application/x-stanford-json`

spec/requests/create_dro_spec.rb
peetucket: as above, I think this is the content type of the wrapper, the blob is the one missing the content type

peetucket: for example, see https://github.com/sul-dlss/sdr-client/blob/main/lib/sdr_client/deposit/upload_files.rb#L51 and https://github.com/sul-dlss/sdr-client/blob/main/lib/sdr_client/deposit/create_resource.rb#L48-L51 for where we post to sdr-api with `application/json` type as the header (with the body being a blob with the separate content-type we are munging here)

openapi.yml
```diff
@@ -29,7 +29,12 @@ paths:
     post:
       tags:
         - files
-      summary: Creates a new file resource
+      summary: |
+        Creates a new file resource.
+        Note that the 'content_type' of the file posted will be set to 'application/octet-stream' if not supplied.
+        Additional values in 'content_type' after a semicolon will be removed (e.g. ';v15' is removed in 'application/text;v15')
+        You may post JSON files with 'content_type' of 'application/x-stanford-json' to avoid 400 errors if posted JSON file is invalid.
+        The system will automatically change the content type to 'application/json' after upload.
```
jmartin-sul, line 37: nice!  i love the text, but i'm pretty sure it belongs on the `/v1/disk/{id}` route below, in the `put` section, as that's where the actual 400 error was coming from:  https://app.honeybadger.io/projects/77112/faults/96225134

i think the `@url=` part of the faraday response is the URL to which the request was made:
```
RuntimeError: unexpected response: #<Faraday::Response:0x00007f58b55fd808 @on_complete_callbacks=[], @env=#<Faraday::Env @method=:put @request_body=#<File:/data/h2-files/jy/fi/jyfi2xh06cszo21unpzmha8tis9m> @url=#<URI::HTTPS https://sdr-api-prod.stanford.edu/v1/disk/eyJfoobarbaz> @request=#<Faraday::RequestOptions timeout=900, read_timeout=1800> @request_headers={"User-Agent"=>"Faraday v2.7.6", "Content-Type"=>"application/json", "Content-Length"=>"216381", "Authorization"=>"Bearer eyJhbGciOiJIUzI1NiJ9.eyJ1c2VyX2lkIjozLCJleHAiOjE2ODY3MTM5NjB9.gKIVYRD4tDhdhWFTw6enHrdcWroREgCcR81wtwevfIA"} @ssl=#<Faraday::SSLOptions verify=true> @response=#<Faraday::Response:0x00007f58b55fd808 ...> @response_headers={"date"=>"Tue, 13 Jun 2023 03:39:23 GMT", "server"=>"Apache/2.4.41 (Ubuntu)", "cache-control"=>"no-cache", "x-request-id"=>"f374e9f7-fe04-416e-91c3-7e444f3b4c44", "x-runtime"=>"0.002156", "x-powered-by"=>"Phusion Passenger(R) 6.0.12", "strict-transport-security"=>"max-age=31536000; includeSubDomains", "status"=>"400 Bad Request", "connection"=>"close", "transfer-encoding"=>"chunked", "content-type"=>"application/json"} @status=400 @reason_phrase="Bad Request" @response_body="{\"id\":\"bad_request\",\"message\":\"Request body wasn't valid JSON.\"}">>
```

per this flow described in the README, i think things are going wrong on step 3: https://github.com/sul-dlss/sdr-api#sequence-of-operations

spec/requests/create_dro_spec.rb
jmartin-sul: oh, yeah, i think you're right!  sorry, thanks for the explanation.

spec/requests/create_dro_spec.rb
jmartin-sul: yup, i think you're right, thanks for the explanation!

jmartin-sul (changes requested):
thanks for the openapi.yml edit, and for rolling in the tests @edsu experimented with for reproducing the undesired 400 in #586!  had a note to myself to add the former and to try doing the latter 🎉 

requesting changes for a small thing: i _think_ the openapi comment needs to be moved to the `/v1/disk` route.  no good deed, and all that 😬

openapi.yml
peetucket: I moved the openapi documentation comment to the `/v1/disk/{id}` route

jmartin-sul (approved):
🌮

PR #601, opened 2023-08-15
Excise references to Symphony

mjgiarlo:
# Why was this change made? 🤔

This commit removes references to Symphony now that the migration to Folio is imminent.

# How was this change tested? 🤨

N/A

### DIFF
--- a/config/initializers/honeybadger.rb
+++ b/config/initializers/honeybadger.rb
@@ -1,8 +1,0 @@
-# frozen_string_literal: true
-
-Honeybadger.configure do |config|
-  config.before_notify do |notice|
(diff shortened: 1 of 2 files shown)


ndushay (approved):
Looks good.  Do we really not need a honeybadger config??

mjgiarlo:
@ndushay 💬 
> Looks good. Do we really not need a honeybadger config??

Nope! Most of our projects don't have an HB config in Rubyland. It's only needed in projects where the config can't live in configuration, like what we had before in sdr-api. Now we'll lean on `config/honeybadger.yml` in shared_configs or ENV values in puppet.

PR #614, opened 2023-09-12
Update cocina models

justinlittman:


openapi.yml
```diff
@@ -239,7 +239,7 @@ paths:
                   exp:
                     type: string
                     description: When the token will expire
-                    format: date-time
```
justinlittman, line 242: The format that is used does not conform with date-time. This is now strictly enforced.

mjgiarlo (approved):

PR #622, opened 2023-10-05
Relax validation of cocina version.

justinlittman:
So that minor cocina gem changes can be rolled out without have to update everywhere.

Unit


ndushay (approved):
great thinking.  Thank you!

PR #663, opened 2024-04-03
Change versioning in UpdateJob.

### LINKED ISSUE #661: UpdateJob should optionally open a version before updating cocina / writing files
The version should only be opened if there is not already an open version.

In addition, the version should be closed once updating is complete (instead of the current logic in `version_or_accession`).

justinlittman:
closes #661

So that UpdateJob is using the latest approach to opening / closing versions.

Unit, integration tests

### DIFF
--- a/app/jobs/update_job.rb
+++ b/app/jobs/update_job.rb
@@ -5,7 +5,7 @@
 # Above that, it will exit and provide the error message in the background_job_result.
 class UpdateJob < ApplicationJob
   queue_as :default
-  attr_accessor :start_workflow
+  attr_reader :background_job_result
 
   # Note that deciding when to stop retrying is handled below. Hence, not providing additional retry configuration
   # for Sidekiq.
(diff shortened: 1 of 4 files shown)


app/jobs/update_job.rb
```diff
@@ -14,42 +14,23 @@ class UpdateJob < ApplicationJob
...
-      Honeybadger.notify("#{error_title}: #{error_detail}",
-                         { external_identifier: existing.externalIdentifier,
-                           current_version: existing.version,
-                           provided_version: model.version })
-      background_job_result.output = { errors: [title: error_title, detail: error_detail] }
-      background_job_result.complete!
-
-      return
-    end
+    existing_version_status = object_client.version.status
```
justinlittman, line 31: Use versioning status from the new endpoint.

app/jobs/update_job.rb
```diff
@@ -14,42 +14,23 @@ class UpdateJob < ApplicationJob
...
+    return unless check_versioning(model, existing_version_status.version)
+    return unless open_new_version_if_needed(model, version_description, existing_version_status, object_client)
```
justinlittman, line 33: Open when we start ...

app/jobs/update_job.rb
```diff
@@ -59,12 +40,10 @@ def perform(model_params:,
 
     background_job_result.output = { druid: model.externalIdentifier }
 
-    versioning_params = { description: version_description || 'Update via sdr-api' }
-
     StageBlobs.stage(signed_ids, model.externalIdentifier)
     StageGlobus.stage(globus_ids, model.externalIdentifier)
 
-    version_or_accession(object_client, model, existing, versioning_params)
+    object_client.version.close
```
justinlittman, line 46: Close when we're done ...

aaron-collier (approved):

PR #711, opened 2024-10-14
Allows specifying whether to close version upon update.

### LINKED ISSUE #710: Support start_workflow for UpdateJob
The IngestJob supports a start_workflow param that determines whether an object is closed after deposit. However, for the UpdateJob there is no ability to control whether the object is closed; it is always closed after deposit.

To support updating a draft, a start_workflow param should be supported for the UpdateJob.

justinlittman:
closes #710

To allow an application to update a draft (without closing).

Unit

### DIFF
--- a/app/jobs/ingest_job.rb
+++ b/app/jobs/ingest_job.rb
@@ -11,7 +11,7 @@ class IngestJob < ApplicationJob
   # @param [Hash] model_params
   # @param [Hash] filename, signed_ids for the blobs
   # @param [BackgroundJobResult] background_job_result
-  # @param [Boolean] start_workflow if true, start accessionWF
+  # @param [Boolean] accession if true, closes the current version
   # @param [Boolean] assign_doi if true, adds DOI to Cocina obj
   # @param [String] priority ('default') determines the relative priority used for the workflow.
   #                                      Value may be 'low' or 'default'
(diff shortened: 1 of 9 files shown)


app/jobs/ingest_job.rb
```diff
@@ -54,8 +54,8 @@ def perform(model_params:, background_job_result:, signed_ids: {}, globus_ids: {
     StageGlobus.stage(globus_ids, druid)
 
     # Close the version, which will also start accessioning
-    # start_workflow will be false in the case of reserving a druid
-    Dor::Services::Client.object(druid).version.close(user_versions:) if start_workflow
+    # close will be false in the case of reserving a druid
```
jmartin-sul, line 57: for consistency, should `start_workflow` have changed to `accession` instead of `close`?

jmartin-sul (approved):
one nitpicky question, but approving.  if you think the suggestion is worthwhile, feel free to just make the change and merge, since it's just a comment.

app/jobs/ingest_job.rb
justinlittman: Great question. I initially changed it to "close", but decided to go with "accession" because it aligns with the API.

That being said, all of SDR API would benefit from a variable name refactor.

PR #713, opened 2024-10-21
Okcomputer check for workflow server

peetucket:

We have outages today with sdr-api-prod being unable to connect to workflow-server-rails.  The errors that resulted from this were not obvious -- and I spent a couple hours chasing down dead ends before landing on this conclusion.  I realize we've talked in the past about not having okcomputer checks that test our own internal services, as nagios should already do this.  But in this case, nagios showed no problems, and an alert like this would save a lot of time to help quickly identify the problem.

Screenshot 2024-10-21 at 3 02 31 PM

QA

### DIFF
--- a/config/initializers/okcomputer.rb
+++ b/config/initializers/okcomputer.rb
@@ -31,6 +31,18 @@ class TablesHaveDataCheck < OkComputer::Check
   end
 end
 
+# make sure we can hit the workflow service
+class WorkflowServerCheck < OkComputer::Check
+  def check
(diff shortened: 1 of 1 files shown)


edsu (approved):
I think this is a good idea. Should we test that it can talk to DSA too?

jmartin-sul (approved):

jmartin-sul:
> I think this is a good idea. Should we test that it can talk to DSA too?

seems reasonable, could be a good small ticket for the prod priorities board, if peter doesn't have the time right now to do a follow up to this PR

PR #731, opened 2025-01-16
Normalize logging practices across services

### LINKED ISSUE #178: Review logging
We're producing huge amounts of logs. This should be reviewed to verify that we are not logging unnecessarily verbosely. Use the standard Rails logger instead of STDOUT to get logs in `log/production.log` and use the `lograge` gem to patch controllers to log less verbosely.

Also, it would be good to update the DevOpsDocs to reflect where logs are going: https://github.com/sul-dlss/DevOpsDocs/blob/master/projects/sdr-api/operations-concerns.md#how-do-i-check-the-logs

mjgiarlo:
# Why was this change made?

To normalize logging practices across services. To do this, lean on lograge and remove duplicate code added on upgrade to Rails 7.1.

Fixes #178 (when paired with sul-dlss/puppet#11735)

Closes #597

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -41,11 +41,6 @@ Rails.application.configure do
   # Force all access to the app over SSL, use Strict-Transport-Security, and use secure cookies.
   config.force_ssl = true
 
-  # Log to STDOUT by default
-  config.logger = ActiveSupport::Logger.new($stdout)
-                                       .tap  { |logger| logger.formatter = Logger::Formatter.new }
-                                       .then { |logger| ActiveSupport::TaggedLogging.new(logger) }
-
   # Prepend all log lines with the following tags.
   config.log_tags = [:request_id]
 
--- a/config/initializers/lograge.rb
+++ b/config/initializers/lograge.rb
@@ -2,4 +2,6 @@
 
 Rails.application.configure do
   config.lograge.enabled = Rails.env.production?
+  config.lograge.base_controller_class = ['ActionController::API', 'ActionController::Base']
+  config.lograge.ignore_actions = ['OkComputer::OkComputerController#show']
 end


config/environments/production.rb
```diff
@@ -41,11 +41,6 @@
   # Force all access to the app over SSL, use Strict-Transport-Security, and use secure cookies.
   config.force_ssl = true
 
-  # Log to STDOUT by default
-  config.logger = ActiveSupport::Logger.new($stdout)
-                                       .tap  { |logger| logger.formatter = Logger::Formatter.new }
-                                       .then { |logger| ActiveSupport::TaggedLogging.new(logger) }
-
```
mjgiarlo, line 48: We didn't need to add this because this file already has https://github.com/sul-dlss/sdr-api/pull/731/files#diff-da60b4e96eff2b132991226d308949e23f4ef3aad45ad59edd09cbc32cc6251eR76-R80 which only lots to `STDOUT` if the `RAILS_LOG_TO_STDOUT` env var is turned on.

justinlittman (approved):

config/initializers/lograge.rb
```diff
@@ -2,4 +2,6 @@
 
 Rails.application.configure do
   config.lograge.enabled = Rails.env.production?
+  config.lograge.base_controller_class = ['ActionController::API', 'ActionController::Base']
```
mjgiarlo, line 5: Patch both sdr-api's controllers and the DirectUploadsController to log less verbosely.

config/initializers/lograge.rb
```diff
@@ -2,4 +2,6 @@
...
+  config.lograge.ignore_actions = ['OkComputer::OkComputerController#show']
```
mjgiarlo, line 6: Do we need to see each GET request Nagios sends to the OkComputer controller every minute?

PR #753, opened 2025-04-15
Bump cocina-models version with openapi.yml changes

lwrubel:
For digital serials work.


justinlittman (approved):

lwrubel:
Will be replaced with an updated new PR.

PR #767, opened 2025-05-29
[HOLD] Upgrade rails 8

aaron-collier:


db/structure.sql
```diff
@@ -360,13 +353,12 @@ ALTER TABLE ONLY public.active_storage_attachments
...
+('20221116163116'),
 ('20220711202210'),
-('20221116163116');
-
+('20210105175540'),
+('20210105175539'),
+('20200514172809'),
+('20200110154606'),
+('20191115013418'),
+('20191114204827');
```
mjgiarlo, line 363: Hmm, I wonder if this change is unrelated to the Rails bump. I wouldn't expect a Rails upgrade to care about the Postgres structure, but then... I could be way off-base.

aaron-collier: Huh, I did not even notice that... Odd... wonder why these old ones etc are getting injected. I'll pull t his out of the PR and investigate separately.

mjgiarlo (approved):

